PR #3, opened 2022-04-07
Adding eslint twilio and prettier config

jhunter-twilio:


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
johnfischelli, line 1: Typically this file is gitignored.

PR #6, opened 2022-06-21
Fix crash when mainTimeZone attribute missing

dremin:
Fall back to the local machine timezone if we have nothing else to use.


jhunter-twilio:
LGTM!

PR #22, opened 2022-07-20
Auto select callback

dremin:
Added setting to control the feature as well.

When a reservation goes away, Flex calls `SelectTask` with task set to `null`. To work around this, here when we place a call back we save the reservation SID in state so that we can override the aforementioned `SelectTask` and provide our own SID.

Also moved `MessageListItem` from `actions` to `components`


jhunter-twilio (approved):
Great job!

PR #25, opened 2022-07-22
Add Flex UI 2.0 version of caller-id feature

dremin:
Other items of note:

- Updated package.json to match plugin builder dependency versions
- Added paste-elements to flex-hooks to allow features to override Paste styles. Included example in code comment for now since no features are using this.
- Not currently using RTK `createAsyncThunk` due to issues using it with Flex, so the old Redux promises middleware is still used.
- Updated custom `WorkerAttributes` interface to be named `CustomWorkerAttributes` and extend flex-ui's `WorkerAttributes`


plugin-flex-ts-template-v2/src/FlexTSTemplatePlugin.tsx
```diff
@@ -33,13 +34,14 @@ export default class FlexTSTemplatePlugin extends FlexPlugin {
...
       RegisterFlexNotifications,
+      CustomizePasteElements,
       CustomizeFlexActions,
       CustomizeFlexComponents,
       CreateSdkClientInstances,
       TeamFilters,
       CustomChatOrchestration,
       SalesForceHooks,
-      cssoverrides
+      CssOverrides
```
jhunter-twilio, line 44: haha i love this

plugin-flex-ts-template-v2/src/types/task-router/Worker.ts
```diff
@@ -21,6 +21,7 @@ export default interface Worker extends EventEmitter {
   workspaceSid: string;
 }
 
+// flex-ui also defines this interface, but doesn't include our custom attributes.
```
jhunter-twilio, line 24: Im wondering if this interface should then extend the existing interface and just add our custom attributes.  So we use CustomWorkerAttributes or something?

jhunter-twilio (approved):

PR #34, opened 2022-08-17
Updating v2 template for scrollable activities

jefferyyates:
Added scrollable-activites to the v2 plugin.
Updated flex-config/index.js to accommodate NOT having any preexisting skills.


jhunter-twilio (approved):
Yeah this looks.much better, thanks for following up with product team

PR #37, opened 2022-08-26
Feature v2 activity reservation handler

dahamilton-twilio:
Ported from V1 folder. 
Changed sdk reservation event handler to new flex events handlers.
Switch component to function component with paste


jhunter-twilio (approved):
looks good

PR #40, opened 2022-08-30
Modifying Chat transfer to before and invoking events

brianschmitt:
Created a new Chat Transfer Action
This action is now invoked from the before Transfer Event.


jhunter-twilio (approved):
looks good!

PR #42, opened 2022-09-02
Adding enhanced crm container to v2

jhunter-twilio:


plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/custom-components/IFrameCRMContainer/IFrameCRMContainer.tsx
```diff
@@ -0,0 +1,15 @@
+import IFrameComponent from "./IFrameCRMComponent";
+import { connect, ConnectedProps } from 'react-redux'
+import { AppState } from '../../../../flex-hooks/states'
+
+
+const mapStateToProps = (state: AppState) => {
+  return {
+    tasks: state.flex.worker.tasks
```
dremin, line 8: You can avoid this entire file by using the new `useFlexSelector` hook to access Flex state in the component.

You will need to convert the component to be a function instead of a class.

Here is an example of using the hook:
```
const tasks = useFlexSelector((state: AppState) => state.flex.worker.tasks);
```

Then in the component you use `tasks` directly rather than accessing via `this.props.tasks`.

PR #70, opened 2022-10-25
Upgrade to flex ui 2.0.0-beta.2

dremin:
### Summary

Several changes had to be made due to beta 2 differences:
- Types are now required when using Actions framework and some other areas. I set these to `any` for now since that is effectively current state anyway.
- Conferences have changed quite a bit, presumably due to Sync dependency removal.
  - I'm noticing that usually `conferenceSid` is undefined when it shouldn't be. As a result I had to change everything to use `task.attributes.conference.sid` instead. Tracking as FLEXEXP-1118.
- `Ticker` component is no longer in flex-ui-core, so I recreated the functionality in our component
- The fake 'connecting' participant when adding an external conference participant seems to go away too quickly now. The churn is unfortunately causing `ConferenceMonitor` to rapidly change `endConferenceOnExit`. Fixed this by creating a new component instead of abusing Flex state.
- Removed FLEXEXP-865 workarounds
- Paste-elements hook required minor changes to work again.

### Checklist


dremin:
Note: Flex 2.0 beta 2 was pulled from being available on hosted Flex; it is currently only available for local development or self-hosting. I don't know if that means we should hang off on merging this.

jhunter-twilio (approved):

PR #76, opened 2022-11-04
Multiple call support

dremin:
### Summary

Is it a hack? Yes. Does it work? Yes!

The idea of this feature is to enable Flex agents already handling a call to be able to handle a second call that gets transferred to them by another worker.

The reason that Flex does not support multiple simultaneous calls out-of-the-box is due to a limitation in the Twilio Voice JavaScript SDK used by Flex. To work around this limitation, the `multi-call` feature instantiates a second Voice SDK `Device` to handle a second incoming call. This works because disabling `allowIncomingWhileBusy` prevents the Voice SDK instance managed by Flex from receiving a second inbound call, allowing our second instance to handle it gracefully.

Due to the Voice SDK limitation, Flex's state maintains an assumption that only one call at a time may be considered active. To work around this, the `multi-call` feature changes the active call in Flex state whenever the selected task changes.

When handling more than one call at once, only one call should be heard by the agent at a time. To handle this, when accepting or un-holding a call, any other active call is immediately placed on hold and muted. Mute is toggled in addition to hold in order to prevent the agent from being recorded on the wrong call.

### Checklist


dremin:
I realized there are a couple more TODO items for some edge cases and different configs:
- Set audioConstraints as configured
- Set acceptOptions as configured
- If initialDeviceCheck is enabled, set the input device before accept.

jhunter-twilio:
A couple of issues
-------	
1. i found transferring multiple times left the participants getting out of sync.  Ultimately transferring 4 times, then on final step hanging up, left the caller on hold with no tasks.
	
2. there is something to consider here around transferring to queue,
- workflow filter not applicable so transfers to queue push to any qualifying agent in the queue which would be anyone with the spare capacity.  We could probably tackle this with an override of the queue transfer screen to be a workflow transfer solution - this is a probably just something to cover in the readme.

jhunter-twilio:
would it be possible to also check the sdk options in the plugin loaded event - and throw a warning notification if the feature is enabled but the sdkOption is set to true?

dremin:
All of the above should be addressed now (except the item that turns out isn't ours!)

jhunter-twilio (approved):
Looks great, fantastic work!

PR #80, opened 2022-11-10
Conference hold workaround

dremin:
### Summary

This branch grew to also contain a few other changes...

1. Added conference hold workaround, controlled by the `hold_workaround` setting. Background: Setting endConferenceOnExit does not take effect when the participant is held, meaning a customer on hold while a third participant is dialed will end the entire conference if they later hang up. The workaround briefly takes them off hold when adding another participant, then puts them back on hold.
2. Added `add_button` setting to the conference feature, which controls the visibility of the "+" button to add a new participant. It also disables some other functionality that is only used for participants added via API. The intention of this setting is to allow the conference feature to be used on accounts with the External Warm Transfer feature enabled. This might be desirable because the conference feature adds the ability for the caller to hang up without ending the conference, as well as a button to kick the caller.
3. Updated to Flex UI 2.0.0-beta.3. This beta fixed some type definitions, which resulted in needing to add a prop to all `IconButton`s. Also reverted the type changes that were made for beta 2.

### Checklist


lhamoudi (approved):
Tested without fix and observed the same issue on v1 (using dialpad-addon plugin) and v2 Flex. I did notice that if customer comes off hold after third participant is added, and then later drops, the conference remains - so there’s a degree of workaround in place already, albeit simplistic. But if customer drops while on hold, the conference does indeed end unceremoniously.

Having this additional level of workaround (and having it be opt-in) is another great value add to customers that might want it. And I’m sure anyone who sees the README and tests it will want to switch it on. Great addition.

**Review**
The code changes behave well in my tests and the code (and cleanup) looks good. You even fixed the additional issue I found when customer remains on hold through the entire agent/external participant interaction, the external participant drops, and then customer drops prior to being brought off hold. Agent previously was left in conference alone - but with the workaround, it is resolved too (because the workaround applies to any endConferenceOnExit update - not just the transition from 2-3 participants).

Only item I spotted was that the 2 new feature flags aren't in `appConfig.js.example` - just ui_attributes. Following discussion, it may merit a separate discussion around removing the duplication that exists in this file. It doesn't seem worth the effort to maintain the 2 files in sync with each other, as the template grows.

jhunter-twilio:
Excellent review Lindsay, thanks! Thorough as ever!

appConfig is only applicable for local environment. That's why it's useful, it means you don't have to deploy anything to the account and test whatever you need locally. Particularly useful for testing someone's PR without interrupting your own environment.

It is a tiny bit cumbersome to maintain both but I think if we add "appConfig updates" to the checklist it's a pretty good trade off for that little overhead

dremin:
I wonder if we do have an opportunity to further simplify the appConfig, given it effectively gets merged with the deployed ui_attributes-- IMO the appConfig should only contain the few settings that the developer wants to be different while running locally, such as the serverless domain or a feature being worked on.

jhunter-twilio:
Excellent point, that makes a lot of sense.

jhunter-twilio:
I think as long as the high level structure is there it should be self evident

dremin:
> I think as long as the high level structure is there it should be self evident

Updated the example appConfigs to have just an example which makes it clear how to use.

lhamoudi:
Nice work

PR #82, opened 2022-11-15
Reduce boilerplate

dremin:
### Summary

Previously each serverless function was mostly boilerplate, making them more difficult to work with and read. This also increased the friction for a developer adding new functions to the project.

Added two wrapper functions that take all of the boilerplate and move it to a single spot:

- `prepareStudioFunction` - validates that required parameters are provided, sets up the response object with the appropriate headers, and provides an error handling function
- `prepareFlexFunction` - the same as above, plus the Flex `TokenValidator`

These wrapper functions are used in the exact same way we typically used `TokenValidator`, wrapping the entire handler function. Now, functions behave exactly the same but are significantly easier to read and write. Here is an example, `get-call-properties`:

```
const { prepareFlexFunction } = require(Runtime.getFunctions()["common/helpers/prepare-function"].path);
const VoiceOperations = require(Runtime.getFunctions()[
  "common/twilio-wrappers/programmable-voice"
].path);

const requiredParameters = [
  { key: "callSid", purpose: "unique ID of call to fetch" },
];

exports.handler = prepareFlexFunction(requiredParameters, async (context, event, callback, response, handleError) => {
  try {
    const { callSid } = event;
    
    const result = await VoiceOperations.fetchProperties({
      context,
      callSid,
      attempts: 0,
    });
    
    const { success, callProperties, status } = result;
    
    response.setStatusCode(status);
    response.setBody({ success, callProperties });
    callback(null, response);
  } catch (error) {
    handleError(error);
  }
});
```

As part of the change, I decided to axe the `scriptName` parameter we were passing around in favor of `context.PATH`. This provides a few benefits:
- `context.PATH` provides the function name and its path in the project structure
- Fewer parameters to pass around to every Twilio wrapper function; we were already passing `context` around
- Developers don't need to remember to name their handler function, which was easy to overlook

### Checklist

### DIFF
--- a/plugin-flex-ts-template-v2/src/feature-library/chat-to-video-escalation/custom-components/SwitchToVideo/SwitchToVideo.tsx
+++ b/plugin-flex-ts-template-v2/src/feature-library/chat-to-video-escalation/custom-components/SwitchToVideo/SwitchToVideo.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { Actions, ITask, Manager } from "@twilio/flex-ui";
+import { Actions, ITask, Manager, ConversationState } from "@twilio/flex-ui";
 import { Flex, Button } from "@twilio-paste/core";
 import { VideoOnIcon } from "@twilio-paste/icons/esm/VideoOnIcon";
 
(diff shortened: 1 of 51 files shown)


jhunter-twilio (approved):
I haven't tested the changes but i've gone through the code and this is a huge improvement in readability.  Excellent job.

One request, if you have not already done so, can you make sure the retry handling works correctly?

dremin:
> One request, if you have not already done so, can you make sure the retry handling works correctly?

It looks like retry handling is working the same as it is in the main branch.

lhamoudi:
Tested without fix and observed the same issue on v1 (using dialpad-addon plugin) and v2 Flex. I did notice that if customer comes off hold _after_ third participant is added, and then later drops, the conference remains - so there’s a degree of workaround in place already, albeit simplistic. But if customer drops while on hold, the conference does indeed end unceremoniously.

Having this additional level of workaround (and having it be opt-in) is another great value add to customers that might want it. And I’m sure anyone who sees the README and tests it will want to switch it on. Great addition.

`npm ci` is taking an age to run for me, so I'll test the new code and complete review tomorrow :)

mschmitt19 (approved):

lhamoudi:
Deleted my old comment which was for PR80 not 82 :)

PR #85, opened 2022-11-17
Fix video breakage from beta 3 type changes

dremin:
### Summary

I changed this file in #82 because it was broken by beta 2, but the types changed again with beta 3 which re-broke it.

### Checklist


mschmitt19 (approved):
Good catch, thanks for updating

PR #88, opened 2022-11-21
Add programmable chat transfer for 2.0

dremin:
### Summary
We already have CBM chat transfer, but customers migrating from 1.x to 2.0 will likely still be running programmable chat when they upgrade, so this feature fills that gap. This is brought over from 1.x unchanged except for break-fixes (not much point in putting significant effort into deprecated functionality). Also updated the video button to match the transfer button style.

The new CBM transfer feature has been renamed to `conversation-transfer` and this feature takes the `chat-transfer` name to allow easy migration from the 1.x feature.

I also fixed some bugs outside of the feature that I ran into while regression testing:
- Fixed update-task-attributes saving null values
- Fixed chat-transfer sending null string to complete-task-for-transfer
- Fixed conversation-transfer when running locally

### Checklist


jhunter-twilio:
Yeah I think I like the last option the best too. Although it might be easier to complete after finalizing the review of multi participant chat.

johnfischelli:
Might I propose `Flex Programmable Chat Transfer` or alternatively `Legacy Chat Transfer` and `Flex Conversations Transfer`? I'm not sure any customers speak in terms of `cbm` specifically. Additionally, our docs refer to `Flex Conversations` https://www.twilio.com/docs/flex/conversations so I feel like this is how customers will think about it and the language they will use.

johnfischelli:
ahhh after looking at the code I see the Flex UI literally has a TaskHelper method `isCBMTask` - so maybe that language is ok - though its super confusing it does not match our docs.

dremin:
> Yeah I think I like the last option the best too. Although it might be easier to complete after finalizing the review of multi participant chat.

Good point, I'll keep this parked until that one is merged.

dremin:
While testing, I found an existing issue where error messages from the create-transfer-task function do not get sent to the frontend. Per the original message in the PR, I'm going to leave the issue as-is for now; we can always revisit if it looks like prog chat will stick around for some reason.

dahamilton-twilio (approved):
looks good to me

PR #92, opened 2022-11-23
Adding WorkerDirectoy - Queue Tab Replacement template 

cherylmj01:
Added feature custom-transfer-directory which enables the replacement of the worker transfer directory and the queue transfer directory enabling the following behavior

- worker - TODO
- queue
  - render different transfer icons for chat channel
  - enable the use of real time data to 
      - present queue insights such as # of tasks or # agents available
      - other data available such as queue wait time but further work required to show it on screen
      - filter out queues that have no agents available
  - provide an improved starting point for augmenting queue transfer list with custom data (imagine the need to filter queues based on skills required to transfer to those queues)
  - provide the ability to enforce queue filters by worker
  - provide ability to enforce global queue filter to filter out system queues.


jhunter-twilio:
@cherylmj01 - I've revised the PR if you want to take a look.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/flex-hooks/notifications/CustomTransferDirectory.ts
```diff
@@ -0,0 +1,21 @@
...
+  {
+    id: CustomTransferDirectoryNotification.FailedLoadingInsightsClient,
+    content: StringTemplates.FailedToLoadInsightsClient,
+    type: flex.NotificationType.warning,
+  },
+  {
+    id: CustomTransferDirectoryNotification.FailedLoadingInsightsData,
+    content: StringTemplates.FailedToLoadInsightsData,
+    type: flex.NotificationType.warning,
+  },
```
dremin, line 20: Given that these are warnings and not critical, a timeout of 3-5 seconds might be nice here.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/flex-hooks/paste-elements/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  TRANSFER_DIR_QUEUE_HORIZONTAL_ROW_CONTAINER: {
+    paddingLeft: '0.5rem',
+    paddingRight: '0.5rem',
+    color: 'inherit',
+    background: 'none',
+    border: 'none',
+    outline: 'none',
+    borderStyle: 'none',
+    borderWidth: '0px 0px 0px 0px',
+    // '-webkit-box-align': 'center',
```
dremin, line 50: Is this needed any more?

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const isFeatureEnabled = (): boolean => {
+  return enabled;
+};
+
+export const isCustomQueueTransferEnabled = (): boolean => {
+  return isFeatureEnabled() && queueEnabled;
+};
+
+export const isCustomWorkerrTransferEnabled = (): boolean => {
```
dremin, line 25: Keyboard thing happened! But, this should be removed until the worker directory is implemented anyway.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueDirectoryTab.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const updatedQueues = transferQueues.current
+      .filter((queue) => {
+        const searchString = searchInputRef.current?.value.toLocaleLowerCase() || '';
+        return queue.name.toLocaleLowerCase().includes(searchString);
+      })
+      .filter((queue) => {
+        if (showOnlyQueuesWithAvailableWorkers()) {
+          // returning only queues with available workers
+          // or queues where meta data is not available
+          return queue.total_available_workers === null || queue.total_available_workers > 0;
```
dremin, line 201: In my account, only the queues that have not seen action in a while have the null value here, so I'm wondering if it makes sense to include those when this setting is used. Maybe if there is an insights client issue, then display the null ones?

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+            ) : (
+              <CallOutgoingIcon
+                key={`queue-item-cold-transfer-icon-${queue.sid}`}
+                decorative={false}
+                title="Warm Transfer"
+              />
+            )}
+          </Button>
+        </Tooltip>
+      </ButtonGroup>
```
dremin, line 112: There are a few scenarios where the warm transfer icon should not render:
- The warm transfer feature has been disabled (or was never enabled in the first place) in Flex Admin
- Conversation transfer has a config option for warm transfer. If that option is disabled, the warm transfer icon should be hidden for CBM tasks.

For the conversation-transfer feature item, [this code](https://github.com/twilio-professional-services/flex-project-template/blob/main/plugin-flex-ts-template-v2/src/feature-library/conversation-transfer/flex-hooks/actions/ShowDirectory.ts) is responsible for hiding the button. Best case scenario is we are able to use css class names such that the existing hiding code works for this as well, but if that isn't possible then checking the config might be needed.

I see a `isWarmTransferEnabled` prop getting passed down here but it is unused. It should probably be used here and then set higher up depending on the above items.

plugin-flex-ts-template-v2/src/types/task-router/Worker.ts
```diff
@@ -36,6 +36,9 @@ export interface CustomWorkerAttributes extends WorkerAttributes {
   // caller-id feature
   selectedCallerId: string;
 
+  // custom-transfer-directory featurer
```
dremin, line 39: Typo!

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/types/ServiceConfiguration.ts
```diff
@@ -0,0 +1,15 @@
+export default interface CustomTransferDirectoryConfig {
+  enabled: boolean;
+  use_legacy_search_icon: boolean;
+  worker: {
+    enabled: boolean;
+  };
```
dremin, line 6: Remove until implemented

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Enable the feature in the flex-config asset for your environment. 
+
+``` javascript
+"custom_transfer_directory": {
+  "enabled": true, // globally enable or disable the feature
+  "use_legacy_search_icon": true, // use legacy icon (recommended if mixing queue overrided with OOTB overrride)
+  "worker" : {
+    "enabled": false // enable the custom worker tab - not currently developed, still to come
+  },
```
dremin, line 32: Remove until implemented

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueDirectoryTab.tsx
```diff
@@ -0,0 +1,286 @@
...
+        if (event.target != null) {
+          filterQueues();
+          setQueueFiltertimer(null);
+        }
+      }, 300),
+    );
+  };
+
+  // async function to retrieve the task queues from the tr sdk
+  // this will trrigger the useEffect for a fetchedQueues update
```
dremin, line 98: keyboard!

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+            onClick={onColdTransferClick}
+          >
+            {TaskHelper.isChatBasedTask(task) ? (
+              <SendIcon key={`queue-item-cold-transfer-icon-${queue.sid}`} decorative={false} title="Warm Transfer" />
+            ) : (
+              <CallOutgoingIcon
+                key={`queue-item-cold-transfer-icon-${queue.sid}`}
+                decorative={false}
+                title="Warm Transfer"
+              />
```
dremin, line 108: The `title` tags here are wrong, but should probably just be removed because they cause a double tooltip.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueDirectoryTab.tsx
```diff
@@ -0,0 +1,286 @@
...
+            await workspaceClient.fetchTaskQueues({
+              Ordering: 'DateUpdated:desc',
+            })
+          ).values(),
+        ) as unknown as Array<IQueue>,
+      );
+  };
+
+  // async function to retrieve queues from the insights client with
+  // agent availability - it shoould be noted the insights client acts
```
dremin, line 113: keyboard!

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/README.md
```diff
@@ -0,0 +1,47 @@
...
+  "queue" : {
+    "enabled": true, // enable the custom queue tab
+    "show_only_queues_with_available_workers": true, 
+    "show_real_time_data" : true, // tool tup for queues will show real time data instead of queue name
+    "enforce_queue_filter_from_worker_object": true, // when true, if `worker.attributes.enforcedQueueFilter` is present, it will be enforced, otherwise ignored
```
dremin, line 37: I wonder if this should instead not be enabled by default given this attribute isn't present by default.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueItem.tsx
```diff
@@ -0,0 +1,115 @@
...
+          element="TRANSFER_DIR_COMMON_ROW_NAME"
+          as="div"
+          title={queue.name}
+          className="Twilio"
+        >
+          {queue.name}
+        </Text>
+      </Tooltip>
+
+      <ButtonGroup key={`queue-item-buttongroup-${queue.sid}`} element="TRANSFER_DIR_COMMON_ROW_BUTTONGROUP" attached>
```
dremin, line 67: The button group's border radius is causing weird focus half-circles when clicking an item. Maybe this could be fixed in the Paste element hook.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueDirectoryTab.tsx
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    setQueueFiltertimer(
+      setTimeout(() => {
+        // eslint-disable-next-line no-eq-null, eqeqeq
+        if (event.target != null) {
+          filterQueues();
+          setQueueFiltertimer(null);
+        }
+      }, 300),
```
dremin, line 93: Is this basically just a debounce? It might be easier to understand to use the lodash `debounce` function which we already have available for use in the template.

dremin (changes requested):
This is a very nice improvement over the default interface! Some pretty minor items I'd like to see adjusted before merging.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/config.ts
```diff
@@ -0,0 +1,55 @@
...
+  return global_exclude_filter;
+};
+
+export const shouldFetchInsightsData = (): boolean => {
+  return showOnlyQueuesWithAvailableWorkers() || showRealTimeQueueData();
+};
+
+export const useLegactSearchIcon = (): boolean => {
+  return use_legacy_search_icon;
+};
```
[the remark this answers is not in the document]
dremin: Side note: Maybe we could make this a bit more explicit by naming it something like `use_paste_search_icon` (which would flip the values) or `use_flex_search_icon`?

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueItem.tsx
jhunter-twilio: lol good catch

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueItem.tsx
dremin: It looks like the `isWarmTransferEnabled` prop out-of-the-box is set by the warm transfer feature flag, which gets you half way here :)

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/QueueDirectoryTab.tsx
jhunter-twilio: yeah thats exactly whats happening - if the data is missing we dont know if it should be excluded or not, so just include it (makes it easier to resolve if the issue is an inactive queue, because we can push something into the queue)

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/README.md
jhunter-twilio: i think its better to have it enabled by default because the absence of the field on the worker is ignored and simply adding it to the worker will have it evaluated for use so its less overhead in configuring and managing.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/CommonDirectoryComponents.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const LegacySearchIcon = () => {
+  return (
+    <div className="Twilio-Icon Twilio-Icon-Search  css-1j3rlv1">
+      <svg>
+        <path d="M15.742 14.153l3.091 3.091a.417.417 0 11-.589.59l-3.091-3.092a5.833 5.833 0 11.59-.59zm-4.409 1.18a5 5 0 100-10 5 5 0 000 10z"></path>
+      </svg>
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
jhunter-twilio: i think we should get rid of the icons folder and keep each feature encapsulated in its own folder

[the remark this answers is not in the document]
dremin: I like that even better!

dremin (approved):
:tada:

PR #94, opened 2022-11-23
Update Paste and use switch in capacity config

dremin:
### Summary

When porting supervisor-capacity to Flex UI 2.0 I left a TODO in to use the Paste `Switch` control once it was fixed. It is now, so I updated the template's version of Paste and switched the capacity checkbox to the switch.

I also fixed the placeholder first option in the caller-id dropdown being selectable when it shouldn't be (unrelated to the above, but figured I could sneak the fix in).

### Checklist


jhunter-twilio (approved):
lgtm

PR #96, opened 2022-11-23
Suppress Flex error during internal calls

dremin:
Note: Please review #98 first, if that approach seems sensible we can close this instead.

### Summary

When on an internal call, Flex is not aware of the conference state, and will throw an error saying such. This is new behavior after Sync removal. However, the call still works as it did before.

This PR works around the issue by suppressing the notification when on an internal call.

In the future, we could theoretically fix this and other issues with internal-call by taking a more 'native' approach allowing TaskRouter to manage the conference (that way it will be aware of conference state, and maybe even fix transfers). I spent a solid chunk of time looking into this, and have some ideas, but they haven't yet panned out, so I think this PR is a reasonable stop-gap.

I also noticed the `pluginsLoaded` handler wasn't referenced for this feature, so I fixed that too.

### Checklist


jhunter-twilio (approved):
yeah this is a good option for now - excited to hear the other solution design ideas

PR #97, opened 2022-11-28
Fix recording internal calls with dual-channel

dremin:
Note: Please review #98 first, if that approach seems sensible we can close this instead.

### Summary

Because Flex doesn't have conference state for internal calls, the dual-channel recording feature would not see the participants as joined and not begin recording as a result. This fix checks the active call in Flex state rather than task conference state to enable recording internal calls.

### Checklist


dremin:
This doesn't make the pause-recording feature work with internal calls. I can add a separate code path for it, but having 2 code paths for everything isn't very sustainable. I'd like to further pursue an alternate design for internal-call that avoids all of these issues.

I think this PR is still worthwhile to allow recordings with the current implementation, assuming the pause-recording feature isn't a necessity for this.

jhunter-twilio (approved):

PR #98, opened 2022-11-29
New approach for internal-call feature

dremin:
### Summary

Completely new approach to the internal-call feature, allowing TaskRouter to orchestrate the conference natively. This addresses issues with the previous approach:
- Internal transfers now work
- External transfers are shown correctly in UI
- Dual-channel recording now works (without code from #97)
- Pause recording now works
- Flex error no longer needs to be suppressed (#96)
- A single task is reported in Flex Insights
- Can be used when running the template locally

The new approach is simple: Call a "dummy" phone number or SIP address (which immediately disconnects, but does not end the conference) and automatically warm transfer to the desired agent.

### Checklist


dremin:
Still going to pursue this, merging the other fixes for the current approach to allow more time to mature this approach.

dremin:
The current issue with this approach is that it incurs additional cost compared to the previous approach by using a phone number. Attempts to implement this approach without adding cost have not been successful so far.

dremin:
Abandoning this PR; a true long-term solution needs to come from the product level to avoid the issues with this PR.

jared-hunter:
So many dashed hopes!

PR #99, opened 2022-11-29
Minor initial bringup improvements

dremin:
### Summary

- Added readme section for deploying the template without a pipeline
- Fixed issue in populate script when a parameter is not provided
- Updated flex-config env generate script to use account SID and auth token if api key/secret are missing, since those can be used instead
- Made device manager button match dialpad button style as another paste-elements example
- Fixed some minor typos

### Checklist


jhunter-twilio (approved):
awesome stuff, thank you!

PR #101, opened 2022-11-30
Fix schedule manager localization shortcomings

dremin:
### Summary

- Fix 'type' not using localized value
- Added utils/RRuleLanguage.ts for localizing recurrence rule text (copied the English one from [the RRule.js repo](https://github.com/jakubroztocil/rrule/blob/master/src/nlp/i18n.ts))

### Checklist


jhunter-twilio (approved):
lgtm

PR #103, opened 2022-12-01
Update CallbackService.ts

mirobotur:
fixed wrong variable definition

### Checklist


dremin (approved):
I didn't see this function referenced anywhere, but good to get this fixed so that it can be used!

PR #104, opened 2022-12-01
Stellwag

aestellwag:
### Summary
Added a new featureFlag check to the V2 portion of the plugin.  This changes how we check if a feature is enabled throughout our code.  I've updated any reference to custom_data or UIAttributes (where applicable) to the new method.  If we agree on this approach and merged, we can work on adding this to V1 if we feel there is value in doing so.


dremin:
@aestellwag this touches a lot of the same things that I touched in #91. That PR consolidates config fetches within each feature to one file. I like your approach of simplifying how config is fetched, so to minimize merge sadness I plan on copying the helper you added into #91 and adapt the new feature-specific files I added to use it.

jhunter-twilio:
@aestellwag - @dremin merged the critical changes from this PR into his other PR as they overlapped a lot.  The great news is both sets of changes were highly compatible! Closing this PR

PR #113, opened 2022-12-06
Windows updates 2

jhunter-twilio:
### Summary

Switching from `concurrently` to `npm-run-all` to support `npm run start:local:v1` and `npm run start:local:v2` commands in windows.  

Concurrently is doing something in the windows cmd shell that breaks the plugin start.  Shell.js also has the same issue, this is why the `npm run start:local` command still doesn't work as it uses shell.js to launch the process after it dynamically identifies the plugin folder.  As a reminder, this script is used after the template has been renamed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ or (for Flex UI v2.x)
 ```bash
 npm run start:local:v2
 ```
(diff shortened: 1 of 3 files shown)


dremin (approved):
lgtm

PR #115, opened 2022-12-06
Better handle Sync errors in serverless

### LINKED ISSUE #111: Reason is not defined
Will this work?
I tested the same approach and it ends up with 
`failed after 0 retry attempts, status code: 500, message: reason is not defined`

https://github.com/twilio-professional-services/flex-project-template/blob/69d3edea836c7f2e23831dad9cbaabe76b6cc188/serverless-functions/src/functions/common/twilio-wrappers/sync.private.js#L47

dremin:
### Summary

Fixes #111 by not catching errors prematurely. I noticed the plugin component that was using the function also wasn't handling the error so I added a basic error notification to handle it.

### Checklist

### DIFF
--- a/plugin-flex-ts-template-v2/package.json
+++ b/plugin-flex-ts-template-v2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plugin-flex-ts-template-v2",
-  "version": "9.0.38",
+  "version": "9.0.39",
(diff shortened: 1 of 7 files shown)


mschmitt19 (approved):
Looks good! 🤙

PR #119, opened 2022-12-10
Agent automation

jhunter-twilio:
### Summary

Added `agent automation` feature and `supervisor complete reservation` feature.

`agent-automation` - selects based on the first configuration set to match the channel and required attributes.
`supervisor-complete-reservation `- allows supervisor to complete tasks that are in wrap.  considered enabling for accepted tasks but completing voice tasks left the agent call open and it was tricky trying to clear up.

### Checklist


dremin:
Looking good! Some thoughts from my initial once-over:

supervisor-complete-task:
- Messaging tasks don't use TaskOverviewCanvas, so mounting elsewhere could enable the functionality for other channels.
- 'Close task' is a bit misleading given it uses the same icon as the button to close the panel, a different icon/tooltip could provide better clarity

agent-automation:
- Auto wrapup didn't occur if I reloaded the page while on a matching call, or reloaded between ending the call and auto wrapup, until I switched to teams view and then back to the agent view. I'm guessing this is due to the logic being driven by a TaskCanvasHeader component. Maybe using an event like taskWrapup and a reconciliation in pluginsLoaded would be worth exploring?
    - I see a comment in the code about not using the events in case of dropped packets, but if those events are missed, wouldn't Flex not know to put the task in the wrapping state anyway?
    - If it remains a component, making it a function-based one instead of class-based would be nice :)
- I noticed that there is logic to check to see if the task has been accepted before selecting it, what is the reasoning around this?
    - Maybe we could hide the disabled accept/reject buttons if a task matches auto_accept?
- Do we want to support nested task attributes under required_attributes?

jhunter-twilio:
thanks Sam, i appreciate the review

supervisor-complete-task:
 - Thanks for catching this - moved to Supervisor.TaskCanvasHeader
 - yeah great feedback, i changed to a destructive button and elaborated more clearly on the tooltip

agent-automation
- this consistently works without any issues for me, can you re-test and make sure the task is qualifying? - i updated the criteria in the common config to qualify all voice and chat tasks automatically. if you are still seeing the issue lets look together
    - no you can miss the event in a noisy network but refresh and get the state of the task as being in wrapped 
    - this design requires to only load the timer once so the componentDidMount hook is ideal, i think this needs to stay a class unless there is an improved approach.  This currently looks like the best one to me.
 - if you auto select before accepting, you'lll need to select it again after accepting.  If the first select is still processing, the second select can on occasion fail.
   - feels like a nice to have future enhancement - id like to cap scope here
 - i dont think i've ever seen complex criteria here but would be a nice to have - id like to cap scope here

dremin:
> thanks Sam, i appreciate the review
> 
> supervisor-complete-task:
> 
>  Thanks for catching this - moved to Supervisor.TaskCanvasHeader

Is there a reason this now requires supervisor barge/coach to be disabled?

>  yeah great feedback, i changed to a destructive button and elaborated more clearly on the tooltip

Nice! Nit: the button looks a bit wrong because of the way Paste pads around the button, if you replace the `div` with a paste `Box` you can use their spacing tokens to add some padding and make it appear correctly. Might help to set size=small on the button.

> agent-automation
>  this consistently works without any issues for me, can you re-test and make sure the task is qualifying? - i updated the criteria in the common config to qualify all voice and chat tasks automatically. if you are still seeing the issue lets look together

Not sure what was happening earlier, but I'm not able to reproduce now 😄 

>  no you can miss the event in a noisy network but refresh and get the state of the task as being in wrapped

Right, that was why I mentioned using pluginsLoaded as well- it could potentially run the same logic but that would run immediately after a refresh. Regardless, I'm not seeing issues any more with the current approach.

>  this design requires to only load the timer once so the componentDidMount hook is ideal, i think this needs to stay a class unless there is an improved approach.  This currently looks like the best one to me.

A `useEffect` hook with no dependencies should be functionally equivalent to componentDidMount.

>  if you auto select before accepting, you'lll need to select it again after accepting.  If the first select is still processing, the second select can on occasion fail.

You might need to show me what you mean, I tested with this code removed, instead selecting before accepting, and didn't see issues.

plugin-flex-ts-template-v2/src/feature-library/supervisor-complete-reservation/custom-components/SupervisorCompleteReservation/SupervisorCompleteReservation.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const completeReservation = async () => {
+    setIsOpen(false);
+    dispatch(Actions.updateReservation(taskSid, reservationSid, "completed"));
+ };
+
+  return (
+    <Box
+      paddingRight="space20"
+      paddingTop="space20">
```
dremin, line 34: This is still showing as cut off for me on the left side. This works for me:
```
padding="space20"
paddingBottom="space0"
```

plugin-flex-ts-template-v2/src/feature-library/agent-automation/flex-hooks/events/taskReceived.ts
```diff
@@ -0,0 +1,50 @@
...
+      } else if (attempts > 10) {
+        clearInterval(selectTaskTimer);
+        attempts = 0;
+      } else {
+        attempts++;
+      }
+    }, 500);
+  } else if (taskConfig.auto_select) {
+    Flex.Actions.invokeAction("SelectTask", { sid });
+  }
```
dremin, line 41: I'd like to learn more about why this is needed, I wasn't able to reproduce any issues by only running SelectTask then AcceptTask, so I must be missing something!

dremin (changes requested):
This is looking good, just a very minor fixup on the button and looking for some more details around the SelectTask logic (we can handle that over zoom or slack)

plugin-flex-ts-template-v2/src/feature-library/agent-automation/flex-hooks/events/taskReceived.ts
jhunter-twilio: looks like using awaits works now 👍

dremin (approved):
LGTM!

PR #131, opened 2023-01-06
Populate defaults in appConfig.js by postinstall

dremin:
### Summary

Currently, following the readme instructions results in everything being disabled. This fixes that by copying the defaults from ui_attributes.common.json into appConfig.js so that the defaults are consolidated in one place in the repo, while keeping the local separation of config.

### Checklist


dremin:
> This looks great, the only thing is the `features` value populated in the appConfig is a json object but it should be a javascript definition

Yeah, this isn't great. I tried doing some processing to remove the unnecessary quotes, but it ended up being fragile. I'll look for some other solutions.

PR #135, opened 2023-01-21
Validate input before adding conference participant

dremin:
### Summary
Adds an error message if the "phone number" field is empty on submit. Previously this would send an empty string to the serverless function and cause errors.

### Checklist


jhunter-twilio (approved):
Looks great 👍

PR #136, opened 2023-01-21
ParameterValidator empty string check

dremin:
### Summary
Currently, ParameterValidator checks if parameters are provided, but doesn't check if they are empty or not. The thought behind this change is that if a parameter can be provided but empty, it probably shouldn't be included in the requiredParameters array anyway. Otherwise, each function will need to add duplicate checks for if each value is empty or not.

Leaving in 'question' status for now to collect feedback and double-check that other functions do not rely on the current behavior.

#135 exposed this issue.

### Checklist


jhunter-twilio:
This makes sense to me now, I can't think of any reason it would not be backwards compatible with any existing behavior/code.

dremin:
Will require updates to conversation-transfer and supervisor-barge-coach functions since it looks like they have some items marked as required that should not be.

dremin:
Tested to confirm regressions are fixed.

PR #138, opened 2023-01-27
Removing flexFlowSid as a required param (as this could be called from TwiML app). And catering to Transcription Callback URL events as the invokers of this function (i.e. slightly different parameter naming convention)

lhamoudi:
### Summary

Added support for creating a voicemail task off the back of a transcribed recording (via Transcription Callback URL). Also made `flexFlowSid` non-mandatory for usage of the `create-callback` function - since it could well be invoked from a TwiML app too. README edited accordingly - to include the alternative transcription-driven usage of the function.

### Checklist


jhunter-twilio:
I see what youre trying to do it makes sense, i wonder though, the reason the flexFlowSid was there was for debugging and tracking, being able to correlate which studio instance triggered the callback etc.  Is there a similar solution we can do from the twiml app and we can change "flexFlowSid" to something like, "callingIdentifier" or something....  open to suggestions

lhamoudi:
I didn't remove the field. Just made it not required. I can amend the
README to better reflect that, and indicate the value in passing it. Not
sure what you would get from a TwiML app - it'd need to be custom generated
I suppose - but yeah I agree a more generic field name would be better. Can
have a think..

lhamoudi:
Will redo with a little bit more thought, and on a branch.

PR #139, opened 2023-01-27
Voicemail Improvements to Cater to Transcription Callback Invocations

lhamoudi:
### Summary

Catering to Transcription Callback events being the invokers of `create-callback` function (i.e. slightly different parameter naming convention). Updated README to mention this alternative approach for initiating the function call via Transcription Callback URL on the Record Voicemail widget.

Removing `flexFlowSid` as a required param from both (Studio an Flex) `create-callback` functions - as this API could be called from a TwiML app instead of a Studio Flow. 

Updated README to reflect optional `flexFlowSid`, and give description as to how it's used.

### Checklist
- [x] README updates

### DIFF
--- a/plugin-flex-ts-template-v2/package.json
+++ b/plugin-flex-ts-template-v2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plugin-flex-ts-template-v2",
-  "version": "9.0.49",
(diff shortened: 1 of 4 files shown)


jhunter-twilio (approved):
tested, works as documented! 👍

PR #154, opened 2023-02-28
Feature discovery

dremin:
### Summary

The goal of this PR is to remove 🍝 such that adding and/or removing features from the template does not require modifying any files outside of the feature itself. Previously, feature flex-hooks were referenced directly in a top-level flex-hooks directory, and feature-specific types were referenced directly in top-level type definitions. Neither of these things are true any more, removing a large barrier to template development, and further reducing possible merge conflicts.

To accomplish this, we now use a `glob` webpack loader to automatically import each feature's `index.ts` file to the top-level template, using the new feature-loader.

A small amount of required boilerplate has been added to the `index.ts` file for each feature to facilitate flexible feature loading, taking configuration into account. This is not an issue now that we have the `add-features` script. The following items were added:
- One line auto-imports all modules in the feature's `flex-hooks` directory.
- A new named export, `register()`, has been added to every feature's `index.ts` file, and the feature-loader calls this two-line function. If the feature is enabled, this function returns a `FeatureDefinition` object, which consists of the feature name and the array of flex-hooks imported.

This provides the feature-loader a mapping of feature names and flex-hooks for all features in the plugin. The feature-loader then logs enabled features and each of their hooks to the browser console, providing the same level of visibility that we had previously. In the future, we could add additional ways to get this output.

Each flex-hook provided to the feature-loader is then checked for specific named exports. These named exports are specific to each hook type, which tells the feature-loader what type each flex-hook module is. The feature-loader will perform a different action specific to the hook type, such as executing functions, adding event listeners, or modifying manager configuration.

Feature-specific types have also been removed from the top-level template type definitions. Each feature can still use its own types, providing type safety at the feature level.

The add-features and remove-features scripts have both been updated to account for the changes above.

Templates/samples for every type of flex-hook have been added to the plugin readme file, along with a description of every flex-hook type.

### Checklist
- [ ] Updated README


dremin:
@dahamilton-twilio @jhunter-twilio @johnfischelli @mschmitt19 @prasanthbpillai @vinitdave27 As you can see, this is a pretty major change to the template. I'm looking for any and all feedback, so please be critical! :)

plugin-flex-ts-template-v2/src/feature-library/chat-transfer/flex-hooks/jsclient-event-listeners/conversations-client/conversationJoined.ts
```diff
@@ -2,15 +2,13 @@ import * as Flex from '@twilio/flex-ui';
 import { getWorkerFriendlyName } from '../../../utils/serverless/ChatTransferService';
 import { Worker } from 'types/task-router';
 import { Conversation } from '@twilio/conversations';
-import { isFeatureEnabled } from '../../../index';
-
+import { FlexJsClient } from "../../../../../types/feature-loader";
 
+export const clientName = FlexJsClient.conversationsClient;
+export const eventName = "conversationJoined";
```
mschmitt19, line 8: Would it make sense to create enums for the different event types for each JS client for consistency? Instead of manually defining the event have something like `ConversationsClient` defined in the `types/feature-loader` directory then reference them similarly to the other enums (`export const eventName = ConversationsClient.conversationJoined`)

plugin-flex-ts-template-v2/src/feature-library/supervisor-capacity/config.ts
```diff
@@ -0,0 +1,12 @@
+import { getFeatureFlags } from '../../utils/configuration';
```
mschmitt19, line 1: General question about the `config.ts` convention -- why do some features have a `config.ts` at their root folder but others don't? It looks like the code to check if the feature is enabled (`if (!isFeatureEnabled()) return {};`) is sometimes in `index.ts` and other times in the `config.ts`.

plugin-flex-ts-template-v2/src/feature-library/chat-transfer/flex-hooks/chat-orchestrator/completed.ts
```diff
@@ -0,0 +1,12 @@
+import * as Flex from "@twilio/flex-ui";
+
+export const chatOrchestratorHook = (flex: typeof Flex, manager: Flex.Manager) => ({
+  event: 'completed',
```
mschmitt19, line 4: Following the conventions in this PR, should the different events for the `chatOrchestrator` also have their own enum to reference the different events? Might be too granular, but thought I'd bring it up

plugin-flex-ts-template-v2/src/feature-library/conversation-transfer/index.ts
```diff
@@ -17,3 +17,8 @@ export const isColdTransferEnabled = () => {
 export const isMultiParticipantEnabled = () => {
   return enabled && multi_participant;
 };
+
+export const register = (): FeatureDefinition => {
+  if (!enabled) return {};
```
mschmitt19, line 22: Should we change this `enabled` check to match the syntax in the other feature libraries (i.e. creating the `isFeatureEnabled()` function and performing the check with `if (!isFeatureEnabled()) return {};`?

mschmitt19 (approved):

plugin-flex-ts-template-v2/src/feature-library/chat-to-video-escalation/flex-hooks/actions/CompleteTask.ts
```diff
@@ -1,11 +1,12 @@
...
+export const actionEvent = FlexActionEvent.before;
+export const actionName = FlexAction.CompleteTask;
+export const actionHook = function beforeCompleteVideoEscalatedChatTask(
   flex: typeof Flex,
   manager: Flex.Manager
 ) {
-  if (!isFeatureEnabled()) return;
 
   flex.Actions.addListener(
     "beforeCompleteTask",
```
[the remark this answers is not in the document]
dremin: Agreed! I'll implement this.

plugin-flex-ts-template-v2/src/feature-library/chat-transfer/flex-hooks/jsclient-event-listeners/conversations-client/conversationJoined.ts
dremin: Good idea, I left it open because these are from non-flex SDKs, but these should be pretty straightforward to define. Maybe we can even import some enums from the SDKs themselves.

plugin-flex-ts-template-v2/src/feature-library/supervisor-capacity/config.ts
dremin: This was to prevent cyclical imports when a flex hook uses config (index.ts imports all flex-hooks, so a flex-hook importing index to get config makes a big sad). I'll make this more consistent; we should establish the pattern up front so others dont run into issues.

plugin-flex-ts-template-v2/src/feature-library/chat-transfer/flex-hooks/chat-orchestrator/completed.ts
dremin: I'm only aware of wrapup and completed, will try to find others and define an enum for it.

plugin-flex-ts-template-v2/src/feature-library/conversation-transfer/index.ts
dremin: Yeah, not sure how this one ended up different.

PR #168, opened 2023-03-23
Post feature discovery fixes

dremin:
### Summary

While testing the eslint PR I uncovered some issues that were caused by the feature-discovery PR:

- FlexAction enums had some incorrect casing, preventing UnholdParticipant and WrapupTask hooks from working
- The built-in Sync client was never ported to the new hooks structure (for registering the tokenUpdated hook). Since it is not a feature, I referenced it explicitly in feature-loader.

### Checklist


jhunter-twilio (approved):
good catch - tested chat transfer and conferencing, both look good.

PR #165, opened 2023-03-22
Add eslint to v2 plugin

dremin:
### Summary

Re-enables eslint for the v2 plugin. This was already in place for the v1 plugin, but was never brought to the v2 plugin.

- Used twilio style as a base
- Turned off non-risky and annoying rules to prevent unnecessary friction
- CI adds comments and annotations for errors and warnings. Only errors prevent merging.
- Used `--fix` to auto-fix what could be (mostly `prettier`), then made manual fixes for the remaining errors and warnings. Left the TODO warnings intentionally.
- None of the code changes should change functionality.

### Checklist
- [x] Tested changes end to end
- [x] Added PR Label "ready-for-review"
- [x] Requested one or more reviewers for the PR

PR #177, opened 2023-04-03
Add eslint to serverless-functions

dremin:
### Summary

Basically the same thing as #165 but for the serverless-functions package:

- Used twilio style as a base
- Turned off non-risky and annoying rules to prevent unnecessary friction
- CI adds comments and annotations for errors and warnings. Only errors prevent merging.
- Used `npm run lint:fix` to auto-fix what could be (mostly `prettier`), then made manual fixes for the remaining errors and warnings.
- None of the code changes should change functionality.

The largest code changes were in the refresh-stats function, which is unused in the Flex plugin. It should behave the same, but extra testing would be appreciated.

### Checklist

### DIFF
--- a/.github/workflows/check_serverless.yaml
+++ b/.github/workflows/check_serverless.yaml
@@ -1,0 +1,64 @@
+name: check_serverless
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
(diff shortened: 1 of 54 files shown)


jhunter-twilio (approved):
another excellent addition!

PR #187, opened 2023-04-12
Handle errors when multi-call is enabled

dremin:
### Summary

The Voice SDK throws errors when the network connection goes away. Previously, this would crash Flex if the network disconnected while on a call and multi-call is enabled. Adding an error handler prevents the crash.

### Checklist


jhunter-twilio (approved):
lgtm!

PR #188, opened 2023-04-13
Create recording improved client feedback

jhunter-twilio:
### Summary

Improving error code logging for create-recording both in serverless and in the front end

### Checklist


serverless-functions/src/functions/common/twilio-wrappers/retry-handler.private.js
```diff
@@ -47,7 +53,7 @@ exports.retryHandler = async (error, parameters, callback) => {
     return callback(updatedParameters);
   }
   console.error(
-    `retrying ${context.PATH}.${callback.name}() failed after ${logWarning}, status code: ${status}, message: ${message}`,
+    `retrying ${context.PATH}.${callback.name}() failed after ${logWarning}, status code: ${status}, message: ${message}, twilioErrorCode: ${twilioErrorCode}, twilioDocPage: ${twilioDocPage}`,
```
dremin, line 56: I presume `twilioErrorCode` and `twilioDocPage` will be null for all errors except those from `create-recording`-- could we exclude those from the error string if they aren't present? (Ideally we'd update all of the call sites to provide those vars, but I'm guessing that's far past the intended scope of this PR)

dremin (approved):

PR #195, opened 2023-04-18
Fix queueNoWorkerDataFilter display by using Redux

dremin:
### Summary

Per the feature readme, the queueNoWorkerDataFilter actually is a facade and it is just setting skill filters under the hood. Because of this, Flex clears out the queue filter component when applying filters, which is a pretty confusing UX.

This fixes that by storing the selected queue in Redux, and then applying it when Flex clears the component.

Also disabled the agent skills filter by default, as it conflicts with the queueNoWorkerDataFilter if both are enabled (due to the queue filter also setting skill filters). It can be enabled instead of queueNoWorkerDataFilter, or can also be enabled in tandem with a more robust queue filter solution (i.e. queueWorkerDataFilter).

This hack seems to work a bit better than I was expecting, so please give it a spin to make sure I didn't miss anything 😄 

### Checklist

### DIFF
--- a/flex-config/ui_attributes.common.json
+++ b/flex-config/ui_attributes.common.json
@@ -86,7 +86,7 @@
           "queue_no_worker_data": true,
           "queue_worker_data": false,
           "team": true,
-          "agent_skills": true
+          "agent_skills": false
         },
         "department_options": [
           "General Management",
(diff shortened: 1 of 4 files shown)


jhunter-twilio:
it's definitely looking a lot better.  Im not sure if this is a new bug or not but when you use a queue that has an OR'd expression, a notification is posted that the queue filter will be removed and ignored because its not supported.  This doesn't behave as intended, instead it seems to return to whatever was selected last, which on the first instance is blank so it returns to blank but on the second instance is stuck on the queue its supposed to remove  Alternatively i you first select. a working queue then a disqualified queue, it goes back to the working queue.

jhunter-twilio (approved):

PR #197, opened 2023-04-18
Dynamic max capacity in omni-channel-capacity-management feature

dremin:
### Summary

Before setting the capacity to 1, we now save the current configured capacity in local storage. This is then used to restore the capacity rather than the hard coded setting of 2. If local storage is missing, a configurable fallback value is used.

Also added a config option for the affected channel while at it.

### Checklist


jhunter-twilio (approved):
lgtm

PR #198, opened 2023-04-18
Fix canned responses with programmable chat

dremin:
### Summary

Currently, accepting a programmable chat task while canned responses are enabled crashes Flex due to a missing attribute. This fixes that and allows canned responses to be used with programmable chat.

### Checklist


mschmitt19 (approved):
Good catch!

PR #199, opened 2023-04-19
Canned responses improvements

dremin:
### Summary

- The crash fix from #198 was incomplete; it would still crash on the second task handled. That should now be fully resolved.
- Added a really basic cache to the canned responses service so that it doesn't need to hit the function for each task. This seemed like a simple shoe-in, let me know if you'd like it done differently or if there was reasoning around the lack of a cache that I'm not aware of.
- Disable/hide canned responses in wrapup mode when they cannot be sent

### Checklist


mschmitt19 (approved):
I like the cache approach -- nice job!

PR #200, opened 2023-04-20
Now deriving workflow SID (for task lookup) from the voice queue

lhamoudi:
Voice Queue friendlyName is the Workflow SID, so no need to explicitly pass this in from Studio Flow anymore. 

Small bugfix too (missing TaskrouterOperations definition).

### Summary

### Checklist
- [ ] Any config changes added to all environment files
- [ ] For Flex features added, ensure the feature list on the README on the project root folder is updated
- [ ] Completed README template for new features
  - [ ] Feature summary - _high level summary of what the feature does_
  - [ ] Flex User Experience with animations - _animations with descriptions if necessary of the user experience in Flex_
  - [ ] Setup and dependencies - _description of any setup steps required for this feature_
  - [ ] How does it work? - _description of plumbing supporting feature_

### DIFF
--- a/serverless-functions/src/functions/features/callback-and-voicemail/studio/wait-experience.protected.js
+++ b/serverless-functions/src/functions/features/callback-and-voicemail/studio/wait-experience.protected.js
@@ -4,6 +4,7 @@
  *
  */
 const VoiceOperations = require(Runtime.getFunctions()['common/twilio-wrappers/programmable-voice'].path);
+const TaskRouterOperations = require(Runtime.getFunctions()['common/twilio-wrappers/taskrouter'].path)
 const CallbackOperations = require(Runtime.getFunctions()['features/callback-and-voicemail/common/callback-operations']
   .path);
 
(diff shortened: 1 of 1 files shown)


lhamoudi:
Correct - that's the Workflow SID for sending the Callback/VM task to (by
default). In this case however, we need the Workflow SID of the *enqueued
call task *- to allow us to locate it using our more scalable approach.

lhamoudi:
It was Terence's suggestion - to alleviate the need to supply the task's
workflow SID via the Hold Music URL - and I figured I'd implement it while
fixing another bug.

lhamoudi:
If anyone can approve this today or tomorrow, I can get it over to my customer :)

dremin:
> If anyone can approve this today or tomorrow, I can get it over to my customer :) 

I can take a look once the lint issues are addressed.

serverless-functions/src/functions/features/callback-and-voicemail/studio/wait-experience.protected.js
```diff
@@ -154,12 +155,13 @@ exports.handler = async (context, event, callback) => {
...
   }
 
-  const { Digits, CallSid, enqueuedWorkflowSid, mode, enqueuedTaskSid, skipGreeting } = event;
+  const { Digits, CallSid, QueueSid, mode, enqueuedTaskSid, skipGreeting } = event;
 
   switch (mode) {
     case 'initialize':
       // Initial logic to find the associated task for the call, and propagate it through to the rest of the TwiML execution
       // If the lookup fails to find the task, the remaining TwiML logic will not offer any callback or voicemail options.
+      const enqueuedWorkflowSid = await VoiceOperations.fetchVoiceQueue({ context, queueSid: QueueSid }).friendlyName;
```
dremin, line 164: A few things here:

- This function needs an `attempts: 0` param passed into it so the retry handler works properly. (Side note, we should probably fix this requirement separately)
- We need to wrap the function call and the await in parentheses so that the friendlyName property isnt accessed until the promise resolves (currently this returns undefined as a result)
- afaict, fetchVoiceQueue's return object has a queueProperties member that holds friendlyName, not on the top level of the return object (currently this returns undefined as a result)

Here's how I got this working:
```
const enqueuedWorkflowSid = (await VoiceOperations.fetchVoiceQueue({ context, queueSid: QueueSid, attempts: 0 }))
        .queueProperties.friendlyName;
```

dremin (changes requested):
Looks good for the most part, just some fixes needed where the returned data is accessed.

dremin (approved):
🎉

PR #203, opened 2023-04-27
Remove need for functions to pass attempts parameter

dremin:
### Summary

A few reasons for this:
- It's unnecessary
- Depending on how the wrapper was written, nothing fails until the retry handler is hit, which makes this wicked easy to overlook (see review comments in #200 for an example)

### Checklist


mschmitt19 (approved):
🚢🚢🚢

PR #204, opened 2023-04-28
Use pluginsInitialized instead of pluginsLoaded

dremin:
### Summary

See FLXUIAPPIN-607. Basically, pluginsLoaded can be emitted before plugins have actually loaded, so our listeners never are invoked!

### Checklist


jhunter-twilio (approved):
lgtm

PR #208, opened 2023-05-02
Automatically fetch current schedule manager config before deploy

dremin:
### Summary

Because schedules are updated via an admin UI, the config file within the repo gets stale, and CI (or local) deployments will override the live config with the stale config. This is bad.

To resolve this issue, a script is run prior to deploy that pulls the currently-deployed asset and replaces the local config file with it.

Because config files are private, a new function was created that allows fetching the config, provided a valid API key and secret for the account. These items are fetched by environment variables if present (such as in CI) or from the local Twilio CLI profile.

### Checklist


jhunter-twilio (approved):
looks good

PR #211, opened 2023-05-04
Auto detect native external warm transfer within conference feature

dremin:
### Summary

Basically this removes the `add_button` setting and replaces it with an automatic detection of the native external warm transfer feature.

### Checklist

### DIFF
--- a/flex-config/ui_attributes.common.json
+++ b/flex-config/ui_attributes.common.json
@@ -29,7 +29,6 @@
       },
       "conference": {
         "enabled": true,
-        "add_button": true,
         "hold_workaround": false
       },
       "enhanced_crm_container": {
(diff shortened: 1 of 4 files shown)


dremin:
> if im understanding this correctly - when xwt is enabled OOTB this feature addresses some bugs? Are there open bugs raised with xwt? - or would the ability for the caller to hang up without ending the conference and the ability to kick the customer participant not being considered as bugs?

The native xwt feature does have several bugs, and my understanding based on the doc I sent via Slack is that the endConferenceOnExit-related gap filled by this feature should be filled by the time it is public. However, I don't believe adding a button to kick the customer is part of that, so this would still fill that gap. However, we may want to adjust this feature to disable completely if the native feature is enabled, once the native feature is more stable.

jhunter-twilio (approved):

PR #213, opened 2023-05-04
Fully transition conversation-transfer to use native transfer directory

dremin:
### Summary
Improves the feature's support with the native transfer directory:

- Enables the native directory to be used when the cold transfer setting is disabled
  - Now shows the 'open transfer directory' button if only multi-participant is enabled
  - Hides the 'cold transfer' button in the native directory if appropriate
- Hides the dialpad tab added to the native directory when the beta native external warm transfer feature is enabled
- Fixes the bug that exists when inviting a queue that fell off the sync index: the invited queue did not show in the participants tab for the first transfer to that queue
  - This was a low-effort fix due to the corner case. It just shows the queue SID, but at least its there and cancel-able.
    - Since this is an edge case, it's probably fine to hit TR to get the queue name. Open to opinions 😄 
  - When the custom worker directory is merged, it could also be affected by the same bug, and this fix covers that too.

Now that the native directory is now fully supported, the invite buttons and modal are removed.

### Checklist


jhunter-twilio (approved):
Looks great

PR #215, opened 2023-05-05
Fix conversation transfer 'leave chat' race condition

dremin:
### Summary

Currently, if both agents click 'leave chat' at the same time, they will both leave the conversation, leaving the customer all alone, preventing future messages from that customer creating a new interaction.

Sad-Pablo-Escobar

This fixes the issue by fetching conversation participants after removing a participant. If only the customer remains, close out the interaction.

Also fixes some other minor bugs:
- If leave chat fails, re-enable the button.
- Fix key warning from React.

### Checklist


jhunter-twilio (approved):
looks great

PR #227, opened 2023-05-15
send Audio File Feature - Story S239

rickmachadotwilio:
### Summary

### Checklist


plugin-flex-ts-template-v2/src/feature-library/send-audio-rec-file/README.md
```diff
@@ -0,0 +1,31 @@
...
+# setup and dependencies
+To use the audio message feature, you need to have access to the Twilio Flex platform. You also need to ensure that your Flex account is set up correctly, and you have the necessary permissions as agent or supervisor.
+
+The audio message feature has the following dependencies:
+
+* A browser that supports audio recording and playback, such as Chrome, Firefox, or Safari.
+* An internet connection with adequate bandwidth to transmit audio messages.
+
+
+# how does it work?
```
jhunter-twilio, line 21: this section is really to describe whats happening under the hood.
something like, the new component is rendered that captures audio as a fine and is pushed to the conversation as a binary file using X api.  The conversation channel is then responsible for pushing that binary file down to each participant on their specific channel interface (whatsapp, sms, webchat) and then the client for that channel is responsible for recognizing the binary file as an audio file and allow the user to play it.... 

or something like that...

dremin:
@rickmachadotwilio 

> 101 ESLint error(s) and 18 ESLint warning(s) found in pull request changed files.

If you run `npm run lint:fix` in the template directory, it should auto-fix several of these issues and will list the remaining issues that you will need to fix manually. Please feel free to reach out to me if you have any questions regarding the linter or how to solve any of the problems it has found! I'll take a deeper look at the PR once those are taken care of.

dremin:
@rickmachadotwilio This feature is awesome, and I can see a lot of use cases for it! I have some suggestions that I believe would improve the experience as well as some problems I encountered during testing.

- To show the recorder, instead of an expand icon above the message input (which is somewhat vague/unclear), a recording icon should be added in the `MessageInputActions` component. You can see the emoji picker feature for an example of this.
- If no recording has been made, the audio player should be hidden
- Since the start/stop recording buttons can't be used at the same time, we should only show one button. It would also be good to add a `title` to the button so that hovering the button shows what it does. It would also be nice if the button was in-line with the audio player instead of below it.
- If the conversation does not support attachments, the button to show the audio recorder should be disabled or hidden. Currently it lets you record but just shows an error afterwards.
- When trying to record another file after reaching the max number of attachments, it is impossible to attach the recording. In addition, after removing an attachment or sending a message, the player still has the recording but it can't be re-attached, which is a bit confusing. I think a good way to solve all of these problems is to stop auto-attaching the recording and instead add an "attach" or "insert" button which will add the recording as an attachment. Once added successfully, the player should go away and the recorder should close.
- The name of the attached file is a bit vague, maybe naming it "voice-recording.mp3" would be more clear?
- I think there was some confusion around how `flex-hooks` works. Right now you're not really taking advantage of them, an explanation is available [here!](https://github.com/twilio-professional-services/flex-project-template/tree/main/plugin-flex-ts-template-v2#flex-hooks)

rickmachadotwilio:
Moving this to a new one.

PR #228, opened 2023-05-15
Remove v1 plugin

dremin:
### Summary

Before this gets merged, we should either create a tag or a branch that can be used to easily access the final v1 code.

- Removed v1 plugin
- Removed references to v1 plugin in readme
- Removed "v2-only" references from readme
- Consolidated start scripts so `npm start` works in all cases including renamed, remove concurrently
- Fix incorrect repo and license in package.json

### Checklist


jhunter-twilio:
this is looking great - only thing to add is do we need the PLUGIN_FOLDER variable in the github actions steps now?

jhunter-twilio (approved):

dremin:
> this is looking great - only thing to add is do we need the PLUGIN_FOLDER variable in the github actions steps now?

For now, yes, as it is used when the plug-in is renamed. However, we could probably simplify this to not be required otherwise.

dremin:
Tag created: https://github.com/twilio-professional-services/flex-project-template/releases/tag/flex_ui_v1

PR #236, opened 2023-05-30
Add Admin UI

dremin:
### Summary

Currently, we really only have a story for an infrastructure-as-code-based approach to configuration of template features. This is perfect for customers planning to use this approach, or who are already using this approach in their other vendor applications. However, this causes a lot of friction for customers who do not desire to use that approach.

The admin-ui feature shows a dynamically-generated UI for managing the deployed flex-config. Each feature can be enabled or disabled, and all config for that feature is available as well, including documentation links. Config can be set globally (flex config api) or only for the current user (worker attributes).

The admin-ui mode of configuration is assumed as default (customers who want to use infrastructure-as-code can simply disable the feature in flex-config). A new env var has been added, `OVERWRITE_CONFIG`, which controls whether or not the flex-config scripts will overwrite config with what is in the repo (when set to `true`), or if it will only deploy net-new config (default).

The interface dynamically generates a form for feature settings based on the deployed config. An appropriate control appears for string, number, and boolean values. Other more complex types, such as arrays or nested objects, show a JSON editor instead to allow editing. Features may provide custom UI to inject to the feature settings form, allowing feature-specific complex configurations. Currently, activity-reservation-handler and schedule-manager have custom admin UI defined.

### Checklist

- [x] Completed README template for new features in `/docs/docs/feature-library/flex-v2` and added feature row entry into the `/docs/docs/feature-library/00_overview.md` file


dremin:
There is an issue: Features which provide custom admin UI do not load the custom UI when the feature is disabled (since the feature isnt loaded!) We could fix this by providing a bespoke template-level method to register the component that doesn't check if it is enabled, will look into options.

jhunter-twilio:
Perhaps your admin feature can follow the same pattern as other feature overlaps, that is

Admin UI feature registers an action
Other features check if admin UI is enabled and fire the action, regardless if they are enabled or not. The payload could contain whatever you need to handle the UI config side.

jhunter-twilio:
Although it just occured to me that may have a loading order issue 🤔

dremin:
> Perhaps your admin feature can follow the same pattern as other feature cross dependencies, that is
> 
> Admin UI feature registers an action Other features check if admin UI is enabled and fire the action, regardless if they are enabled or not. The payload could contain whatever you need to handle the UI config side.

@jhunter-twilio This is more or less the current state. The problem is that the feature-loader will not register the feature's flex-hook for that action because the feature is not enabled. I'm thinking a new entrypoint will need to be added to a feature's index.ts for registering the action that does not check if the feature is enabled.

jhunter-twilio:
> > Perhaps your admin feature can follow the same pattern as other feature cross dependencies, that is
> > 
> > Admin UI feature registers an action Other features check if admin UI is enabled and fire the action, regardless if they are enabled or not. The payload could contain whatever you need to handle the UI config side.
> 
> @jhunter-twilio This is more or less the current state. The problem is that the feature-loader will not register the feature's flex-hook for that action because the feature is not enabled. I'm thinking a new entrypoint will need to be added to a feature's index.ts for registering the action that does not check if the feature is enabled.

I've not looked at the code so I'm pretty ignorant on the implementation and I'm just thinking abstractly. I assumed the Admin UI feature would be self-contained, what you're trying to solve is the discovery of other features. Hence the thought that the information needed by the admin UI off the other features could be found in the payload of the event. As I type this out I realize you're probably talking about the custom components for each feature that are required to be loaded in the admin UI for non-standard config, which are contained in each independent feature, I'm going back to my PTO then 😎

jhunter-twilio:
Or couldn't you pass those along in the event too? Sorry cant help myself, I need a problem to solve 😂

jhunter-twilio:
Just thinking out loud, no need to explain if it's not viable

dremin:
@jhunter-twilio Yeah the hope is to keep everything self-contained so that there are no cross-dependencies. Right now features use an action hook to register their admin component, but that of course only happens if the feature is enabled. I could probably modify feature-loader so that it does different behavior for that specific action, but I think adding special-cases like that is probably a bad direction.

I'm considering making an "admin-hook" or something of the sort that is not loaded by feature-loader. Instead, it is loaded by the admin-ui feature using a glob pattern (similar to how feature-loader discovers features).

jhunter-twilio:
This is excellent, I love the new pattern we landed on and the addition of supporting individual vs global configuration changes.

some thoughts:

- I see the reset to default button - i dont have any strong feelings on this but my first thought was, should this be a global reset to default instead of an individual one.
- for features that have some configuration dependency - should we have a standard notification icon to highlight that?
  - possibly even better, could we use APIs to deduce if the dependencies are in place and block enabling the feature without them?
- Activity reservation handler has the activities hard coded that you can select for each function.  I'm not sure this is the right approach, the consumer can really nominate whatever activity they want for each behavior - we just created some sample ones in the feature to clearly illustrate the activity state switching.
- the callback docs link is currently broke, likely due to the recent renaming of the feature.
- maybe at a future date we could render the UI slightly differently, show casing a single gif of how the feature behaves as part of the component card. This could be a nice addition.

jhunter-twilio (approved):
Excellent work!

PR #237, opened 2023-05-30
Make activity-skill-filter happy in Flex UI 2.2.0

dremin:
### Summary

Flex.Activity requires the `disabled` prop in 2.2.0. However, 2.1.1 and earlier do not define that prop in the type interface. Add the prop and disable type checking for the file to allow usage across versions in spite of the incompatible interface.

### Checklist


mschmitt19 (approved):
🚢

PR #240, opened 2023-06-01
Use task/worker attributes within canned responses

dremin:
### Summary

When specifying the text of a canned response, you may now include task and/or worker attributes via template variable substitution. To do so, use the format `{{task.attribute_name_goes_here}}` or `{{worker.attribute_name_goes_here}}` within the response text. If the attribute exists, the template variable is replaced with the contents of the attribute. Nested attributes are also supported, such as `{{task.customers.name}}`.

### Checklist


jhunter-twilio (approved):
works great!

PR #247, opened 2023-06-16
Addressing issue with chat-to-video feature

jhunter-twilio:
### Summary

some feedback from @leroychan that the use of chat-to-video had some hiccups

- the localization effort had encoded the url 
  - fixed
- the deployment did not include the chat-to-video assets and some manual work arounds were used
  - the assets have unique ids in their filenames so the deployment scripts were updated to auto generate these and deploy as part of serverless as well as ensure when running locally everything lines up with the local serverless environment.


serverless-functions/.gitignore
```diff
@@ -1,2 +1,2 @@
 # VIDEO. IGNORE THE BUILT CUSTOMER FACING APP ASSETS
-src/assets/features/chat-to-video
\ No newline at end of file
+#src/assets/features/chat-to-video
```
dremin, line 2: Rather than committing built assets, maybe the build should be part of CI?

leroychan: @dremin If we can already pre-built the "video-app-quickstart" web-app into the serverless functions as an asset, that would be even better.

jhunter-twilio: yeah absolutely, this is what i was going to do i was just capturing Leroys changes before i parked it to do something else last Friday

web-app-examples/video-app-quickstart/.env.example
```diff
@@ -1,3 +1,8 @@
-NEXT_PUBLIC_SERVERLESS_FUNCTIONS_DOMAIN=https://xxxxx
-ACCOUNT_SID=ACxxxxxxxxxx
-AUTH_TOKEN=xxxxxxxxxx
+# Domain name for serverless functions
+# if running locally this should be http://localhost:3001
+NEXT_PUBLIC_SERVERLESS_FUNCTIONS_DOMAIN=<PLACEHOLDER_SERVERLESS_DOMAIN>
+
+## Uncomment and configure the following only if using the webapp standalone
+## these are not needed if using the assests deployed to serverless-functions
```
dremin, line 6: assets :)

scripts/common.js
```diff
@@ -269,14 +314,14 @@ exports.populateFlexConfigPlaceholders = function populateFlexConfigPlaceholders
...
-      if(serviceFunctionsDomain){
-        shell.sed('-i', /<PLACEHOLDER_SCHEDULE_MANAGER_DOMAIN>/g, `${scheduledFunctionsDomain}`, configFile);
-      }
+  if(shell.test('-e', configFile)){
+    if(serviceFunctionsDomain){
+      shell.sed('-i', /<PLACEHOLDER_SERVERLESS_DOMAIN>/g, `${serviceFunctionsDomain}`, configFile);
     }
+    if(serviceFunctionsDomain){
+      shell.sed('-i', /<PLACEHOLDER_SCHEDULE_MANAGER_DOMAIN>/g, `${scheduledFunctionsDomain}`, configFile);
+    }
```
dremin, line 323: I know this was just a formatting change but while reading I noticed the variable being checked in the `if` statement is not correct here!

scripts/common.js
```diff
@@ -260,6 +278,33 @@ exports.generateFlexConfigEnv = function generateFlexConfigEnv(context, flexConf
...
+    shell.cp(videoAppEnvExample, videoAppConfig);
+  }
+
+  var {
+    serviceFunctionsDomain
+     } = context
+
+    if(shell.test('-e', videoAppConfig)){
+      if(serviceFunctionsDomain && isLocal){
+        shell.sed('-i', /<PLACEHOLDER_SERVERLESS_DOMAIN>/g, `${localEnvironment}`, videoAppConfig);
```
dremin, line 298: Is there a reason this shouldn't use localhost when no serverless domain was found?

plugin-flex-ts-template-v2/src/feature-library/chat-to-video-escalation/custom-components/SwitchToVideo/SwitchToVideo.tsx
```diff
@@ -77,7 +77,7 @@ const SwitchToVideo: React.FunctionComponent<SwitchToVideoProps> = ({ task, conv
         }
 
         Actions.invokeAction('SendMessage', {
-          body: templates[StringTemplates.InviteMessage]({ videoLink: response.full_url }),
+          body: `${templates[StringTemplates.InviteMessage]()}: ${response.full_url}`,
```
dremin, line 80: I think the `:` should be part of the translate-able string because some languages format that intent differently.

dremin (changes requested):
Looks good! Just a couple mini things.

scripts/common.js
jhunter-twilio: yeah it would create a nonsensical state -the serverrlesss domain should always be found.  its only when we specifically want it to be set to local we set it to local.  Absence of the serverrless domain means something bigger is an issue and setting this to local because of that doesnt accomplish anything.

jhunter-twilio:
experienced some odd behavior with final commits not showing up in PR

scripts/common.js
dremin: Ah, makes sense!

PR #248, opened 2023-06-16
Use serverless to fetch voicemail with HTTP auth

dremin:
### Summary

Given recording media will [require HTTP auth beginning July 31](https://www.twilio.com/en-us/changelog/upcoming-security-changes-enforcing-http-authentication-for-media), we need to support it for retrieving voicemail recordings. This PR adds a new serverless function that will fetch recording media using the recording SID, returning it to the client as base64-encoded data.

The recordingUrl task attribute is no longer used after this change, but I kept it in the docs since it could be useful for other uses. We don't send the recordingUrl to the serverless function because we don't want to allow that function to hit any arbitrary URL using the account SID and auth token, which would be a security risk.

I was concerned about performance, but a request for a 2.5 minute voicemail completes in about 400ms. The serverless function requests the mp3 version of the recording, which is much smaller than the wav version.

### Checklist

### DIFF
--- a/plugin-flex-ts-template-v2/src/feature-library/callback-and-voicemail/custom-components/CallbackAndVoicemail/CallbackAndVoicemail.tsx
+++ b/plugin-flex-ts-template-v2/src/feature-library/callback-and-voicemail/custom-components/CallbackAndVoicemail/CallbackAndVoicemail.tsx
@@ -1,5 +1,5 @@
 import { ITask, useFlexSelector, Manager, Template, templates } from '@twilio/flex-ui';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { DateTime } from 'luxon';
 import { TaskAttributes } from 'types/task-router/Task';
 import { Button, Box, Heading, Text, Flex } from '@twilio-paste/core';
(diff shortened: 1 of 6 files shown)


plugin-flex-ts-template-v2/src/feature-library/callback-and-voicemail/custom-components/CallbackAndVoicemail/CallbackAndVoicemail.tsx
```diff
@@ -60,6 +61,19 @@ export const CallbackAndVoicemail = ({ task, allowRequeue, maxAttempts }: Callba
   const disableCallCustomerButton = disableRetryButton || workerOffline(workerActivitySid);
   const thisAttempt = callBackData?.attempts ? Number(callBackData.attempts) + 1 : 1;
 
+  const [voicemail, setVoicemail] = useState(null as FetchVoicemailResponse | null);
+
+  const fetchVoicemail = async () => {
+    setVoicemail(null);
+    if (callBackData?.recordingSid && !callBackData.isDeleted) {
+      setVoicemail(await CallbackService.fetchVoicemail(callBackData.recordingSid));
```
jhunter-twilio, line 69: it looks like if the fetch fails for whatever reason to return a valid response the screen will show "loading voicemail" indefinitely.  Perhaps we could do a check here on the response so if there is a failure to retrieve the voicemail we can inform the agent that it failed

jhunter-twilio (changes requested):
just a little tweak on handling failed tetch and informing the user - otherwise it looks great to me!

plugin-flex-ts-template-v2/src/feature-library/callback-and-voicemail/custom-components/CallbackAndVoicemail/CallbackAndVoicemail.tsx
dremin: Added a check and an option to retry.

jhunter-twilio (approved):
Works perfect, i love the try again button.

PR #250, opened 2023-06-21
Chat to video localization and deployment fixes

jhunter-twilio:
### Summary
some feedback from @leroychan that the use of chat-to-video had some hiccups

- the localization effort had encoded the url
  - fixed
- the deployment did not include the chat-to-video assets and some manual work arounds were used
  - the assets have unique ids in their filenames so the deployment scripts were updated to auto generate these and deploy as part of serverless as well as ensure when running locally everything lines up with the local serverless environment.


dremin (approved):
LGTM! 🌮

PR #254, opened 2023-06-23
Add cascading menus to canned responses

dremin:
### Summary

Feedback on the MessageInputActions version of canned responses was that the menu was too long to be useful in most cases. This adds submenus for each category instead, which cleans things up.

I moved the menu to be the first button so that there is more room for the menu inside the task canvas without clipping.

### Checklist


jhunter-twilio (approved):
looks great, it would be cool if the menus went more than 1 level deep 👍

PR #255, opened 2023-06-23
Add custom, translate-able Activities filter

dremin:
### Summary

The out-of-box activities filter included with Flex is hard-coded in English. This re-implements that filter using a manger string so that it can be translated.

Changes the teams view filters hook to no longer include the out-of-box activities filter, since now we are providing our own by default. However, if there are no teams view filter hooks defined, then the out-of-box activities filter is used.

### Checklist


jhunter-twilio (approved):
👍

PR #257, opened 2023-06-23
Require admin role to update global config

dremin:
### Summary

The admin-ui feature is only available to users with the admin role. This adds protection to the update-config serverless function so that malicious agents cannot use it.

### Checklist


jhunter-twilio (approved):
Nice safety catch

PR #259, opened 2023-06-23
Fix error with empty sync doc

dremin:
### Summary

Fixes errors when supervisor-barge-coach has not been used before. The reason is this value is `undefined` rather than `null`, which is what was being checked before.

### Checklist


aestellwag (approved):
Tested and looks good, good catch!

PR #265, opened 2023-06-27
Work around Flex bug with parking

dremin:
### Summary

When parking an interaction in Flex, sometimes a `CONVERSATION_UPDATE` Redux action is emitted after a `CONVERSATION_UNLOAD` action. This results in bad state in Flex UI, which means that if the interaction is un-parked during the same Flex UI session, and the same agent receives the conversation, it will not load properly until Flex UI is reloaded.

This hack works around this bug successfully by emitting `CONVERSATION_UNLOAD` 2 seconds later. It was originally conceived for conversations orchestrated outside the Flex Interactions API, but it turns out it is needed for interactions too.

### Checklist


jhunter-twilio (approved):
Oh yeah this was bugging me when I was testing! 😁

PR #267, opened 2023-06-27
Retry using exponential backoff with full jitter

dremin:
### Summary

Adjusted the included ApiService to use a better retry algorithm when it receives a 429. In most cases this is handled by the retry handler within serverless, but that has limits in place which means we can still receive them client-side. Plus, ApiService can be used to make requests to other services, too. The new algorithm [should drastically reduce](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) the amount of requests made as part of the retry logic.

Also added comment clarifying the response status codes handled within the serverless retry handler.

### Checklist

### DIFF
--- a/plugin-flex-ts-template-v2/src/utils/serverless/ApiService/index.ts
+++ b/plugin-flex-ts-template-v2/src/utils/serverless/ApiService/index.ts
@@ -1,8 +1,11 @@
 import * as Flex from '@twilio/flex-ui';
-import { random } from 'lodash';
 
 import { EncodedParams } from '../../../types/serverless';
 import { getFeatureFlags } from '../../configuration';
+
+const MAX_ATTEMPTS = 10;
+const MAX_RETRY_DELAY = 3000;
+const RETRY_INTERVAL = 100;
 
 async function delay<T>(ms: number, result?: T) {
   return new Promise((resolve) => setTimeout(() => resolve(result), ms));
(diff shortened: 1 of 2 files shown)


serverless-functions/src/functions/common/helpers/retry-handler.private.js
```diff
@@ -39,6 +39,10 @@ exports.retryHandler = async (error, parameters, callback) => {
   const retryAttemptsMessage = attempts === 1 ? `${attempts} retry attempt` : `${attempts} retry attempts`;
   const message = errorMessage ? errorMessage : error;
 
+  // Perform retries only on retry-able response status codes
+  // 412: Received upon ETag conflict; retry without delay
+  // 429: Rate limited; retry with delay
+  // 503: Internal error; retry with delay
```
jhunter-twilio, line 45: This should be service unavailable not internal error

plugin-flex-ts-template-v2/src/utils/serverless/ApiService/index.ts
```diff
@@ -1,9 +1,12 @@
 import * as Flex from '@twilio/flex-ui';
-import { random } from 'lodash';
```
jhunter-twilio, line 2: I'm not clear on the merits of removing this rather than just using the max vs min entries when using it.

jhunter-twilio (changes requested):
Happy with the change just curious on the removal of random. Comment update for service unavailable and we are good to go.

serverless-functions/src/functions/common/helpers/retry-handler.private.js
dremin: Description was fetched from here https://www.twilio.com/docs/api/errors/503

plugin-flex-ts-template-v2/src/utils/serverless/ApiService/index.ts
dremin: This was replaced with `Math.random()`. That provides us with the "full jitter" algorithm from [this article](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), which is the same algorithm used by the twilio-node client for exponential backoff.

dremin: As for the merits of full jitter versus a smaller range of jitter, the simulations in the article show that there are actually no benefits when capping the jitter range--the performance is actually better with full jitter.

serverless-functions/src/functions/common/helpers/retry-handler.private.js
jhunter-twilio: But is it not an http status code ?? 😂😂🤦‍♂️

Okay fair enough

jhunter-twilio (approved):

PR #273, opened 2023-06-29
Enhance deploy for any environment

jhunter-twilio:
### Summary

Before, the release scripts were limited to 4 environments - now they are entirely driven by the github environment passed to them containing account sid, api key, api secret.

this creates a less rigid solution where developers can easily all use their own environments as part of their repository

### Checklist


flex-config/index.js
```diff
@@ -1,11 +1,12 @@
 const axios = require("axios");
 const dotenv = require("dotenv");
-const { promises: Fs } = require('fs')
+const fs = require('fs');
```
dremin, line 3: If moving to the sync version of `fs`, we need to refactor the existing calls to `fs` to use that instead of await and try/catch. Otherwise, we should go back to the promises version.

dremin (changes requested):
Everything looks good other than the `fs` change.

dremin (approved):
LGTM! 🌮

PR #279, opened 2023-06-30
Fix changing views while WorkerDirectory is open

dremin:
### Summary

It seems some TaskHelper functions do not like being passed a null task.

### Checklist


jhunter-twilio (approved):
LGTM

PR #280, opened 2023-06-30
Fix freeze from dual-channel feature during accept

dremin:
### Summary

The dual channel feature needs to wait for certain state to be available before starting the recording. Prior to this PR, Flex UI would be unresponsive during this wait, which could be long depending on connection and service availability. This PR does some minor refactoring which allows us to no longer `await` within the `taskAccepted` handler, so that Flex UI remains responsive.

### Checklist


jhunter-twilio (approved):
lgtm

PR #281, opened 2023-07-03
Use documented URL format, reduce API requests

dremin:
### Summary

In some accounts, there is no mediaUrl present when fetching recording details. Therefore, always use the documented recording URL format instead of attempting to fetch the mediaUrl.

Snuck in an unrelated change excluding local .env files from git.

### Checklist


.gitignore
```diff
@@ -182,6 +182,6 @@ override.tf.json
 .terraformrc
 terraform.rc
 
-#ignore local env
-.env.local
-.env.my_hosted_flex
+# Ignore local env(s), allowing only the example env
+.env.*
+!.env.example
```
jhunter-twilio, line 187: This will prevent users from committing overrides to the autogenerated environments for hosted flex deployments though

jhunter-twilio (changes requested):

.gitignore
dremin: Just updated the docs so that users may do this if needed. Per our Slack discussion, hopefully we can work towards eliminating the need altogether :)

jhunter-twilio (approved):
Solid! Good job!

PR #293, opened 2023-07-21
Fix recording after multiple transfers to worker

dremin:
### Summary

Makes sure we record the 'joined' worker participant, not the 'left' worker participant.

### Checklist


jhunter-twilio (approved):
Oh yeah that makes sense, good catch @johnfischelli

PR #297, opened 2023-07-26
Making variables available for terraform studio functions

jhunter-twilio:
### Summary

while doing some studio development recently, discovered there was some overhead still in creating references to functions in the studio flows.  Provisioning variables to be used for services along with two common studio function calls

- serverless - create callback
- schedule manager - check schedule


dremin (approved):
LGTM!

PR #301, opened 2023-07-27
Dynamic feature removal script

dremin:
### Summary

There are a couple use cases, common to existing customers, that the template doesn't currently support well without manual effort:
- Deploying a limited set of features
- Deploying the template for only one of its features

This PR aims to solve that by updating the remove-features script to be much more dynamic. Now it accepts additional parameters on which features to keep when running the script. The following improvements were made to allow this:
- Refactor setup scripts to do better directory checking and share more code. This also involved some changes to the local setup for schedule-manager, which had some vestigial logic that made it differ from the standard serverless package setup
- Removed the `template-files` directories that contained the "no features" files, instead having the script dynamically modify existing files to remove only the undesired features.
- Added comments to workflow YAML such that the script has a way to identify which regions pertain to each feature.
- Added a doc!

### Checklist

- [x] Added docs


jhunter-twilio:
Note - one thing that may be confusing is that the admin ui panel will show the configuration for any features that may have already been deployed to the environment.  You might want to call that out somewhere in thee admin-ui docs

PR #302, opened 2023-07-28
Updating docs

jhunter-twilio:
### Summary
 
further doc updates for activity reservation handler as an example standard.

enhanced configuration management with more standard details


docs/docs/setup-guides/07_configure_the_template.md
```diff
@@ -0,0 +1,89 @@
...
+### config management
+
+#### the custom_data object
+The template maintains the configuration that is deployed to [hosted flex configuration](https://www.twilio.com/docs/flex/developer/config/flex-configuration-rest-api#ui_attributes) in version control under the [`flex-config`](/how-it-works/flex-config) folder.  Here you will find `ui_attributes.common.json` file containing the main configuration set.  
+
+At the time of a GitHub action script deploy of the flex template, when an environment is provided, a new file is generated from `ui_attributes.example.json` and it is called `ui_attributes.<env-name>.json`.  After creating this file any placeholder values in the file are replaced as part of the GitHub deployment scripts.  The contents of this file are merged over the top of the `ui_attributes.common.json` file and the output creates the final configuration set that is pushed to the [hosted flex configuration api](https://www.twilio.com/docs/flex/developer/config/flex-configuration-rest-api#ui_attributes).
+
+#### the data model
+
+The custom_data model that is pushed into ui_attributes follows this pattern
```
dremin, line 27: I think we should mention that we replace hyphens in the feature name with underscores for config.

docs/docs/setup-guides/07_configure_the_template.md
```diff
@@ -0,0 +1,89 @@
...
+
+There are two strategies for managing the configuration set which are mutually exclusive
+
+#### admin ui
+
+You can use the [admin-ui feature](/feature-library/admin-ui), which is the default management style, to manage the configuration.  
+
+:::tip Developer Tip 
+
+Wheen running [locally](#local-configuration), this feature directly ignores what is in `appConfig.js` and shows only what is in [hosted flex configuration](https://www.twilio.com/docs/flex/developer/config/flex-configuration-rest-api#ui_attributes).  As a result this can cause confusion and for that reason is disabled by default when running flex project template locally.
```
dremin, line 53: This isn't completely true. The "myself" config options for admin-ui could be used locally to perform per-worker config overrides; that will take precedence over the appConfig.js or the hosted config. That doesnt change the rationale for disabling it locally, however, per-worker overrides would apply locally even if admin-ui is disabled, so someone could get into a bad place without this info.

docs/docs/setup-guides/07_configure_the_template.md
```diff
@@ -0,0 +1,89 @@
...
+
+::: 
+
+#### version control
```
dremin, line 57: Might be good to use the infrastructure as code term here.

dremin (approved):
🎉

Issue #303, opened 2023-07-30
Agent activity is updated multiple times automatically after changes in activity-reservation-handler plugin


RusinovIG:
After deploying changes from this PR https://github.com/twilio-professional-services/flex-project-template/pull/295 to our account, few agents faced issue, that their activity status was changing automatically for some time (when they were inactive in Flex tab).
One example is: agent went for a lunch before completing wrapup, then went online (at 19:17) and status was changing for 2 minutes automatically:
![image](https://github.com/twilio-professional-services/flex-project-template/assets/1404336/4c8b9701-1ee9-4f8b-9dbe-9b90c8e314d7)
More brutal one, where status was changing for 2 hours from Backup - Online - Backup - Online - Offline every 3 seconds:
https://github.com/twilio-professional-services/flex-project-template/assets/1404336/9615417b-0b6e-4e52-aa49-8b1ae57e0ff1

I'm not sure what other details can I provide? We've seen this problem in Insights afterwards, and not reported immediately by agents.

jhunter-twilio:
@RusinovIG thanks for raising the issue. We'll look into it and get back to you.

In the interim, I might suggest commenting out 

https://github.com/twilio-professional-services/flex-project-template/blob/81f85845c2d8ca5a6200a5168588f77146d82ac4/plugin-flex-ts-template-v2/src/feature-library/activity-reservation-handler/flex-hooks/events/workerActivityUpdated.ts

As this is the likely culprit of the flip flopping. Doing so will mean a Supervisors Activity update will put an Agent in whatever selected state and leave them there but that's probably a lot more preferable over the flip flopping for the time being.

jhunter-twilio:
@RusinovIG do you know if it is common to see agents leaving open multiple flex windows? Or have cause to open multiple flex windows?

Ideally it would be good to hear if you are using 'Enforce Single Session' to rule that out.

RusinovIG:
I don't think agents use / need to use multiple Flex windows, but I'll check if that was the case for those specific agents.
I'm not aware of 'Enforce Single Session' setting and quick search in the docs don't show me anything. @jhunter-twilio Could you point me, where I could see this setting?

Regarding wrapup - Yes, I believe agent just went to lunch without completing wrapup, that's why he wasn't switched to logout immediately, but still you could see status changed twice wrapup <> logout during lunch break (between 18:04 and 19:12).

The main problem however happened at 19:17, when agent went back online.

[...]

PR #304, opened 2023-07-30
Reservation bug 02

jhunter-twilio:
### Summary

issue raised #303 that I am currently unable to reproduce, however given how this manifests the following changes are made that harden the expected behavior.


dremin (approved):
This can only help, fingers crossed it solves the issue for all!

PR #308, opened 2023-07-31
Add schedule check to terraform voice flow

dremin:
### Summary

Adds a schedule check to the Voice IVR deployed by Terraform out-of-box. Uses the same config as the example in the schedule-manager readme.

### Checklist


jhunter-twilio:
So is the idea that everything is in place to lookup the schedule but as it stands it fails?

i like the idea of passing in the "example" to the schedule check so its a little more self evident how you look up a schedule.  What do you think @dremin ?

jhunter-twilio (approved):

PR #321, opened 2023-08-18
Updated Callback to make request for other number

viralgandhi:
### Summary

### Checklist

- [x] Completed README template for new features in `/docs/docs/feature-library/` and added feature row entry into the `/docs/docs/feature-library/00_overview.md` file


dremin (changes requested):
Thanks for working on this @viralgandhi! A couple things to look at:
1. Could you please address the eslint/prettier issues? An easy way to see this locally is to run `npm run lint:fix` (which will auto fix most issues) then run `npm run lint` to see any remaining issues.
2. It looks like the option to choose your number only applies to callbacks and not voicemails. However, voicemails are also a form of callback, so they should also have the option.

viralgandhi:
@dremin , I've fixed issues related to eslint. can you please explain #2? There will not be able changes to voicemail in my view. Thats why we are checking if `_isVoicemail_` request in `handleCallbackOrVoicemailSelected()`

jhunter-twilio:
Hi @viralgandhi 

A voicemail is a type of callback. The agent receives the voicemail and calls the customer back. 

I think @dremin was saying that the customer journey for leaving the voicemail should also include an option to set your callback number.

viralgandhi:
Hi @jhunter-twilio @dremin , Not sure if I understand the usecase here. When end user chooses voicemail option and one is leaving a voicemail saying that 'please call me back on +1xxxxx number' then will it not be duplicate information where you have high chance of getting different number in voicemail vs `numberToCall` field?
We can connect on this to discuss more if needed.

dremin:
> Hi @jhunter-twilio @dremin , Not sure if I understand the usecase here. When end user chooses voicemail option and one is leaving a voicemail saying that 'please call me back on +1xxxxx number' then will it not be duplicate information where you have high chance of getting different number in voicemail vs `numberToCall` field? We can connect on this to discuss more if needed.

I see what you mean--typically a voicemail includes the number to call in the audio portion. However, if the number to dial is not correct in the task attributes, then the agent would need to manually dial the call back to the original caller, which would cause incorrect reporting. However, that is no worse than the current state of things today, so we can look further into what to do later.

serverless-functions/src/functions/features/callback-and-voicemail/studio/wait-experience.protected.js
```diff
@@ -109,32 +115,42 @@ async function handleCallbackOrVoicemailSelected(context, isVoicemail, callSid,
...
 
   const cancelReason = isVoicemail ? 'Opted to leave a voicemail' : 'Opted to request a callback';
-  const mode = isVoicemail ? 'record-voicemail' : 'submit-callback';
+  const mode = isVoicemail ? 'record-voicemail' : 'handle-callback-choice';
 
   const task = await fetchTask(context, taskSid);
 
   // Redirect Call to callback or voicemail logic. We need to update the call with a new TwiML URL vs using twiml.redirect() - since
   // we are still in the waitUrl TwiML execution - and it's not possible to use the <Record> verb in here. We piggyback on the same approach for callbacks,
   // though technically these could be handled entirely in the waitUrl TwiML
```
dremin, line 124: In my opinion, now that the voicemail and callback logic are separate, this function doesn't make sense to handle both any more. Maybe it makes sense to split this function up?

serverless-functions/src/functions/features/callback-and-voicemail/studio/wait-experience.protected.js
```diff
@@ -231,12 +246,83 @@ exports.handler = async (context, event, callback) => {
...
+
+        const gather = twiml.gather({
+          input: 'dtmf',
+          timeout: 15,
+          numDigits: 1,
+          finishOnKey: '#',
+          action: `${domain}/features/callback-and-voicemail/studio/wait-experience?mode=handle-callback-for-other-number-confirmation&enqueuedTaskSid=${enqueuedTaskSid}&updatedPhoneNumber=${Digits.trim()}`,
+          method: 'GET',
+        });
+        gather.say(` press 1 to confirm and 2 to re-enter`);
```
dremin, line 292: Can the message string be extracted into the `options.messages` object to allow easier customizing/localizing?

dremin (changes requested):

viralgandhi:
Hi Sam, the scenario you explained that agent may have to manually dial a
call will occur in any case. And we may not cover all cases for reporting.
May be disposition is one of the way. I would more worry about customer
experience that we are asking customer to punch in their phone number AND
record a voicemail. As a customer I would rather goto callback option then.

Regards,
Viral Gandhi

dremin:
> Hi Sam, the scenario you explained that agent may have to manually dial a call will occur in any case. And we may not cover all cases for reporting. May be disposition is one of the way. I would more worry about customer experience that we are asking customer to punch in their phone number AND record a voicemail. As a customer I would rather goto callback option then. Regards, Viral Gandhi

I agree, for now lets not ask voicemail for the number.

dremin (approved):
🎉

PR #323, opened 2023-08-19
disabling auto accept by default

jhunter-twilio:
### Summary

Per polly vote - disabling auto accept as a default behavior


dremin (approved):
LGTM!

PR #330, opened 2023-08-23
Ring notification for incoming voice and chat

viralgandhi:
### Summary

_Provide a summary of the change_

### Checklist

- [x] Updated documentation


plugin-flex-ts-template-v2/src/feature-library/ring-notification/flex-hooks/jsclinet-event-listeners/worker-client/reservationCreated.ts
jhunter-twilio, line 1: Logically this all looks good but the jsclient-event-listeners is really the legacy approach - we should be using the events hooks.  You can see an example [here](https://github.com/twilio-professional-services/flex-project-template/tree/main/plugin-flex-ts-template-v2/src/feature-library/activity-reservation-handler/flex-hooks/events) 

Given the nature of the code, i can see that breaking things out could be significant and unnecessary complexity.

Could you add a comment in the code just acknowledging that using the manager events is the typically preferred approach but this method makes the code much simpler?

docs/docs/feature-library/ring-notification.md
jhunter-twilio, line 1: Could you enhance the documentation to follow the standard set [here](https://twilio-professional-services.github.io/flex-project-template/feature-library/activity-reservation-handler) and [here](https://twilio-professional-services.github.io/flex-project-template/feature-library/activity-skill-filter)

jhunter-twilio (changes requested):
some documentation updates but otherwise its looking good

docs/docs/feature-library/ring-notification.md
viralgandhi: Thanks for the feedback @jhunter-twilio , I've updated the comment and document. Let me know if anything else is needed.

plugin-flex-ts-template-v2/src/feature-library/ring-notification/flex-hooks/jsclinet-event-listeners/worker-client/reservationCreated.ts
dremin: @viralgandhi There is a typo in the folder name `jsclinet-event-listeners`, could you please fix it to `jsclient-event-listeners`?

plugin-flex-ts-template-v2/src/feature-library/ring-notification/helpers/audioPlayerManagerHelper.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { getFeatureFlags } from '../../../utils/configuration';
+
+class AudioPlayerManagerHelper {
+  _mediaId: string | undefined;
+
+  play = (): void => {
+    const custom_data = getFeatureFlags() || {};
+    this._mediaId = Flex.AudioPlayerManager.play({
+      url: `https://${custom_data.serverless_functions_domain}/features/ring-notification/phone_ringing.mp3`,
```
dremin, line 11: The URL doesn't work when running `localhost`. If you look at the `chat-to-video` feature's `SwitchToVideo` component you can see an example that works in all scenarios. Could you please adjust the logic here to be the same?

dremin (changes requested):
This looks mostly good to go, just a couple small fixes!

dremin (approved):
:tada:

PR #332, opened 2023-08-24
Introducing TR Events Framework & Batch Processing

jhunter-twilio:
### Summary

There are many long running backend operations we want to be able to provide but are currently hampered by the short runtime for twilio serverless functions.  Well not any longer!

This PR introduces a protocol around batch processing that allows us create a array of work items that are then processed in batch. handing results from function to function in a daisy chained manner.

The use case tackled to flush out this protocol was syncing worker objects with an array of their eligible queue sids - which can then be used on the teams view to concretely see which agents are eligible for which queues.

some points to note about this approach 
- worker object has a max 4KB object size, 56 queue sids will take up about half this available space so there is a limit to the success of this approach.
- extensive testing with batch processing has shown it to work very reliably.  I have been testing 5100 agents and the processing takes ~5 minutes, the error handling we have built into all of our functions has paid dividends here.
- in the future i imagine this approach being used for en masse updates of worker skills or anything else that will take a lot of operations 
- the batch processor passes back a sync doc id that can be used to monitor progress on the front end for future use cases.

I've created this draft PR before taking this any further to solicit some feedback or identify any other holes that may require some refactoring before moving to documentation.

For the moment the branch has the workspace event callback url set to the functions but i think we might want to disable that by default for the final output - open to feedback here if you think otherwise.

### Checklist

- [ ] Updated documentation


dremin:
This has been re-homed ➡️ https://github.com/twilio-professional-services/tr-events-batch-processing

PR #337, opened 2023-09-06
✨ S238: Resume Recently Paused/Parked Conversations

Romanti-Ezer:
### Summary

- 🔧 New configuration option `show_menu`
- If `show_menu` is `true`, display the "Recent Interaction List" view
- On parking the conversation, save the conversation status as a Sync Map Item
- ✨ Unpark the conversation from the "Recent Interaction List" view

[Airtable Story Link](https://airtable.com/app5cEnrEEA3sYz5B/pagRnM9J3dCuc4D2D?v6YX7=recdM8OeTI92x7Lkm)

### Checklist

- [x] Updated documentation

### Demo

unpark-interaction


plugin-flex-ts-template-v2/src/feature-library/park-interaction/utils/UnparkInteractionService.ts
dremin, line 1: IMO this can be merged into `ParkInteractionService` as they are ultimately part of the same feature.

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
```diff
@@ -44,6 +46,8 @@ exports.handler = prepareStudioFunction(requiredParameters, async (context, even
         properties: {
           workspace_sid: context.TWILIO_FLEX_WORKSPACE_SID,
           workflow_sid: workflowSid,
+          queue_sid: queueSid,
+          worker_sid: workerSid,
```
dremin, line 50: The routing models for unparking might be different when using the interactions list versus customer-initiated unparking. For example, when unparking from the list I would want the interaction to route to me, but when the customer unparks by sending a reply, they might want it to go to any available worker. Maybe this should be configurable, or maybe unparks from the list should have different routing behavior? (paging @jhunter-twilio)

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
```diff
@@ -3,6 +3,9 @@ const { prepareFlexFunction, extractStandardResponse } = require(Runtime.getFunc
 ].path);
 const ConversationsOperations = require(Runtime.getFunctions()['common/twilio-wrappers/conversations'].path);
 const InteractionsOperations = require(Runtime.getFunctions()['common/twilio-wrappers/interactions'].path);
+const SyncClient = require('twilio-sync').Client;
+
+const syncToken = require(Runtime.getFunctions()['common/twilio-wrappers/get-sync-token'].path);
```
dremin, line 8: Rather than using the Sync client, can we use Sync API via the Twilio client? That will avoid running into Sync client limits.

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
dremin: I also think we should remove the conversation from the Sync map when it is unparked.

plugin-flex-ts-template-v2/src/feature-library/park-interaction/custom-components/ParkView/ParkView.tsx
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    getRecentInteractionList();
+  }, []);
+
+  return (
+    <Box width="100%">
+      <Box paddingTop="space40" paddingLeft="space60">
+        <Heading as="h3" variant="heading30">
+          {templates[StringTemplates.RecentInteractionList]()}
```
dremin, line 79: In another comment I mentioned removing a conversation from the list when it is unparked. That turns this from a 'recent' list to more specifically a 'parked' list. Maybe we should rename this list to Parked Interactions?

plugin-flex-ts-template-v2/src/feature-library/park-interaction/actions/ParkInteraction.ts
```diff
@@ -71,3 +72,17 @@ export const parkInteraction = async (payload: ParkInteractionPayload) => {
...
 };
+
+export const unparkInteraction = async (payload: UnparkInteractionPayload) => {
+  try {
+    await UnparkInteractionService.unparkInteraction(payload.ConversationSid, payload.WebhookSid);
+    return Notifications.showNotification(UnparkInteractionNotification.UnparkSuccess);
+  } catch (error) {
+    const message = (error as any)?.message;
+    console.log(message);
+    if (message.status === 400) return Notifications.showNotification(UnparkInteractionNotification.UnparkError);
```
dremin, line 83: When unparking fails due to a closed conversation, can we remove it from the Sync map?

dremin (changes requested):
Nice work @Romanti-Ezer! Overall this is really good.

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
dremin: If you look at the chat-to-video feature, there is an example of using Sync API via the Twilio client.

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
rickmachadotwilio: 2 cents here:
For this unpark feature routing model, the task/interaction should be routed to same agent.
For customer-initiated unparking that could route to any available agent, as it is today, if the customer wants to change that behavior, they can use something as Known Agent Routing (preferred agent) but they may have to review/change their TR workflow to use it.

flex-config/ui_attributes.common.json
```diff
@@ -249,7 +249,8 @@
         "show_menu": true
       },
       "park_interaction": {
-        "enabled": true
+        "enabled": true,
+        "show_menu": false
```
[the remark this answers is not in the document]
Romanti-Ezer: Agree! I have updated this name.

plugin-flex-ts-template-v2/src/feature-library/park-interaction/custom-components/ParkView/ParkView.tsx
Romanti-Ezer: Makes sense, especially if we're going to remove the interaction after unparked. I've renamed.

plugin-flex-ts-template-v2/src/feature-library/park-interaction/utils/UnparkInteractionService.ts
Romanti-Ezer: Agree, I've merged them

plugin-flex-ts-template-v2/src/feature-library/park-interaction/utils/SyncHelper.js
```diff
@@ -0,0 +1,30 @@
+import { Manager, Notifications } from '@twilio/flex-ui';
+import { SyncClient } from 'twilio-sync';
+
+import { UnparkInteractionNotification } from '../flex-hooks/notifications';
+
+const SYNC_CLIENT = new SyncClient(Manager.getInstance().user.token);
```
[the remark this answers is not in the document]
Romanti-Ezer: Good to know that, just modified to use this shared client!

plugin-flex-ts-template-v2/src/feature-library/park-interaction/actions/ParkInteraction.ts
Romanti-Ezer: Done

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
Romanti-Ezer: I think it makes sense to have it separated:
1. If customer-initiated unparking: go to any available worker
2. If agent-initiated unparking: route to the agent directly

I'll update this function to check the source of unparking, and add `queue_sid` and `worker_sid` conditionally

Romanti-Ezer: Implemented this in this commit [61d39aa](https://github.com/twilio-professional-services/flex-project-template/pull/337/commits/61d39aaf1cc5becafaff1a4e11b15c1fce68b63a)

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
Romanti-Ezer: Refactored this piece to use that sync wrapper, and deleted the `get-sync-token` function that I had created.

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
dremin, line 1: We need to also handle removing from the Sync map when the customer initiates unpark. (For example, when the agent parks, then the customer unparks, currently the interaction is shown in the parked list even though it is not parked, and unparking it in this state causes problems.)

Ideally this would be done in the same place for both scenarios (customer unpark or agent unpark). I think we should set a conversation attribute via the `park-interaction` function that indicates which Sync map the conversation was added to (null if not added to a Sync map, such as when `show_list` is disabled). Then, in this `unpark-interaction` function, we can check that attribute for a Sync map, and remove the conversation from the Sync map if present. Then we could remove the code in the plugin that does the same thing. Let me know if this makes sense! 😅

dremin (changes requested):
I think we need to change how we remove from the Sync map so that it works more consistently; see the below comment 😅

serverless-functions/package.json
```diff
@@ -18,7 +18,8 @@
     "lodash": "^4.17.21",
     "randomstring": "^1.2.2",
     "twilio": "^4.11.0",
-    "twilio-flex-token-validator": "^1.5.6"
+    "twilio-flex-token-validator": "^1.5.6",
+    "twilio-sync": "^3.3.3"
```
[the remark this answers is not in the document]
Romanti-Ezer: Ohh, that's right, I'll remove that!

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
Romanti-Ezer: Yes, this totally makes sense!
I only see benefits from doing this, like keeping Sync clean, avoiding errors and centralizing the removal processing.
I'll work on this.

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
```diff
@@ -106,8 +85,36 @@ exports.handler = prepareFlexFunction(requiredParameters, async (context, event,
...
+        attributes.mapSid = createMapItem.mapSid;
+        attributes.mapItemKey = createMapItem.key;
+      }
+
+      const updateAttributesResponse = await ConversationsOperations.updateAttributes({
+        context,
+        conversationSid,
+        attributes: JSON.stringify(attributes),
+      });
+      if (!updateAttributesResponse.success) throw updateAttributesResponse.message;
```
Romanti-Ezer, line 117: Now updating conversation attributes later
And adding these new attributes if a sync map item was created

docs/docs/feature-library/park-interaction.md
```diff
@@ -7,11 +7,7 @@ title: park-interaction
...
 
-Furthermore, you have the option to enable the 'Parked Interactions' list, which empowers agents to access and manage parked conversations, with a specific focus on Web Chat and WhatsApp. This list allows agents to efficiently unpark conversations directly from their interface.
-- The feature leverages Sync MapItems to store the state of recently parked interactions, with a time-to-live (TTL) of 24 hours
-- Attempting to unpark a closed or failed conversation will result in an error
-- Once attempting to unpark an interaction, the associated Sync MapItem is deleted
-- If the unparking is worker-initiated from the list, `queue_id` and `worder_sid` attributes are added to the request, so the interaction is routed directly to the worker. Otherwise, if customer-initiated, these are not added
+Furthermore, you have the option to enable the 'Parked Interactions' list, which empowers agents to access and manage parked conversations, with a specific focus on Web Chat and WhatsApp. This list allows agents to efficiently unpark conversations directly from their interface. Details [here](#parked-interactions-list-details).
```
Romanti-Ezer, line 10: Moved details to the last section, since this is a summary.

serverless-functions/src/functions/features/park-interaction/common/unpark-interaction.js
Romanti-Ezer: @dremin Implemented this here: [4485364](https://github.com/twilio-professional-services/flex-project-template/pull/337/commits/448536460a13b0ec00f26a09459509512a9dfae8) please let me know what do you think

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
```diff
@@ -46,13 +51,43 @@ exports.handler = prepareFlexFunction(requiredParameters, async (context, event,
...
         status: 'closed',
       });
+      if (!removeAgentResponse.success) throw removeAgentResponse.message;
+
+      let createMapItem = {};
+      if (createUpdateSyncMapItem) {
+        // Open a Sync Map by unique name and update its data
+        const syncMap = await SyncOperations.createMap({
+          context,
+          uniqueName: workerName,
```
dremin, line 68: I think using only worker name as the `uniqueName` could be problematic, because it could conflict with items created by other developers who might be short-sighted. We should add something to the name that indicates that it has something to do with parking. Also, I think we could also remove the need to pass workerName altogether if we use worker SID instead of name!

For example `ParkedInteractions_${workerSid}`

dremin (changes requested):
This is looking very good, thank you for working through the review items! 😎 The only thing left is something I should have noticed earlier but missed (sorry!) Once this is set we are good to go!

serverless-functions/src/functions/features/park-interaction/flex/park-interaction.js
Romanti-Ezer: Just made this change! Totally agree on having a more specified/hard to accidentally duplicate, unique name for the Map. This is one of those things you just pick something while developing and forget to change haha, another good catch, Sam!
And don't worry about requesting changes, I feel it gets more robust every time, and I'm learning a lot, so keep them coming if needed!

plugin-flex-ts-template-v2/src/feature-library/park-interaction/custom-components/ParkView/ParkViewTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const ParkViewTable = (props: ParkViewTableProps) => {
+  const [isUnparkingSid, setIsUnparkingSid] = useState('');
+  const instanceLanguage = Manager?.getInstance().configuration.language || 'en-US';
+
+  const resumeInteraction = async (ConversationSid: string, WebhookSid: string) => {
+    setIsUnparkingSid(ConversationSid);
+    try {
+      await Actions.invokeAction('UnparkInteraction', { ConversationSid, WebhookSid });
```
dremin, line 30: Could we reload the table after this action completes? I noticed that I can keep clicking the unpark button while the notification banner is showing, and that causes me to get multiple tasks for the same conversation.

dremin (changes requested):
I'm sorry, this is the last thing, I swear!! 🙃 Everything else is perfect in my testing!

plugin-flex-ts-template-v2/src/feature-library/park-interaction/custom-components/ParkView/ParkViewTable.tsx
Romanti-Ezer: Just included this! Please take a look when you can.

dremin (approved):
Thanks for working on this and staying with it through the review! LGTM 🎉

PR #339, opened 2023-09-11
Save TF State as assets

ChrisAmora:
### Summary

Move terraform state management to Twilio Cloud function service.

### Checklist

- [x] Updated documentation


infra-as-code/state/pull_state.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+echo "Trying to import Terraform State for $ENVIRONMENT"
+echo "### Job Results "
+
+IFS="|" read -ra tf_state_files <<<"$TF_STATE_FILES"
+
+tfstate_bucket=$(twilio api:serverless:v1:services:fetch --sid tfstate -o json | jq -c '.[].domainBase // empty' | sed 's/"//g')
+
+if [ -n "$tfstate_bucket" ]; then
```
jhunter-twilio, line 10: i realize it complicates things a little more but do you think we could store and retrieve these as protected assets?

.github/workflows/terraform_deploy.yml
jhunter-twilio, line 1: it seems this was renamed from .yaml to .yml - if theres no particular reason do you think you could rename it back?

infra-as-code/state/deploy-function/functions/deploy.js
jhunter-twilio, line 1: would it be easier to just delete the service and get rid of this? 

`twilio api:serverless:v1:services:remove --sid ZS.....`

if that is not possible for whatever reason can we add a comment in here explaining this function is really just used as a method of "deleting" the tfstate by blasting over the top of the service with this function?

jhunter-twilio (changes requested):
@ChrisAmora - this is an excellent bit of work - i am excited to move it into the main branch.  I have some thoughts for minor improvements before we do that.  Please let me know what you think.

infra-as-code/state/pull_state.sh
ChrisAmora: I tried to do that, but I didn't find a way to download the asset.
Here, it says that only Twilio can download a protected asset.
image
I tried to do curl -u "$TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN" -o - $url but it didn't work too :( 
Do you know some other way to do that? Thank you.

.github/workflows/terraform_deploy.yml
ChrisAmora: I just renamed it back!

infra-as-code/state/deploy-function/functions/deploy.js
ChrisAmora: This is the function that I use to deploy the assets. I will document it!

infra-as-code/state/pull_state.sh
jhunter-twilio: followed up directly.

ChrisAmora: FIxed. Bui we will need the TWILIO_AUTH_TOKEN now!

dremin: Is it possible to do this without TWILIO_AUTH_TOKEN? I'm not sure we want to promote using the auth token in CI.

dremin: What if we use a private asset, but use a public function to access it? The public function could accept the API key/secret and validate it before returning the asset data.

dremin: Here is an example of exactly this! 😄 https://github.com/twilio-professional-services/flex-project-template/blob/main/serverless-schedule-manager/functions/admin/fetch-config.js

lvaladares (changes requested):
can we start moving the pipelines to makefile structure? need it to add serverless deploys

ChrisAmora:
> can we start moving the pipelines to makefile structure? need it to add serverless deploys

Not in this PR. The objective here is to change the deployment pipeline as seamlessly as possible.

.github/workflows/flex_deploy.yaml
dremin, line 1: Can you remove the spaces that got added before the `# FEATURE` and `# END FEATURE` lines? These are used by scripting to remove blocks of code and I think this spacing will result in invalid yaml if not removed.

lvaladares (approved):

.github/workflows/flex_deploy.yaml
ChrisAmora: Sure! it was the auto formatter.

ChrisAmora: I didn't know it was used for that too. Good to know.

jhunter-twilio (approved):
this looks great and puts us in a much better spot for improving the "import" stage!! Great work! thank you for the tweaks!

PR #342, opened 2023-09-13
New feature 'sip mute' for muting worker non-webrtc calls

chaosloth:
### Summary

Add a mute button for non-WebRTC calls (i.e. Calls via SIP or PSTN). The default mute button requires the Voice SDK to be used as it calls the local method to mute, this plugin introduces a replacement UI button with a companion serverless function to mute the Flex worker by modifying the conference participants

### Checklist

- [X] Tested changes end to end
- [X] Updated documentation
- [X] Requested one or more reviewers


chaosloth:
@dremin would you be so kind as to review, it's similar to the common serverless functions for conference-participant

dremin (changes requested):
I noticed the 'end call' button is also non-functional with a SIP phone. I think it would make a lot of sense to also address that in this feature (would need a different name), what do you think?

dremin:
Could you also run `npm run lint:fix` to auto-fix and expose any further issues caught by the linter?
`4 ESLint error(s) and 1 ESLint warning(s) found in pull request changed files.`

chaosloth:
TY @dremin  Moving to new feature "sip-support" adding above changes

chaosloth:
New PR with feature #350

PR #343, opened 2023-09-13
Automatically set env vars from GitHub

dremin:
### Summary

When building on the template, there are some use cases where the default set of environment vars should be overridden:
- Creating integrations that require access token / service account details as secrets
  - The developer would add a placeholder for the var in `.env.example`, then the scripting takes care of the rest
- Manually setting a value that was not automatically fetched
  - For example when the mappings file was not updated for a name change
- Temporarily overriding a value for testing purposes

This change essentially passes through the environment secrets and variables defined in GitHub into the runner's environment, causing our existing setup scripts to use them for creating the environment files. The setup script prefers a locally-defined environment var and will skip fetching that var from the API if it would have otherwise.

### Checklist

- [ ] Updated documentation

### DIFF
--- a/.github/workflows/flex_deploy.yaml
+++ b/.github/workflows/flex_deploy.yaml
@@ -60,15 +60,17 @@ jobs:
     environment: ${{ inputs.environment }}
     env:
       ENVIRONMENT: ${{ inputs.environment }}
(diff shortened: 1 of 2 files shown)


jhunter-twilio (approved):
Nice!

PR #344, opened 2023-09-13
Updating the regex logic for queue filter to work

cherylmj01:
### Summary

The initial regex provided for Queue filter under the custom-teams-view feature was not working for expressions where the right hand side would not be in quotes. This would also cause the queue filter to not work because of parsing issues.  
So I updated the code to accomodate for all kind of different expressions so that the queue filter can be applied to the list of agents.

### Checklist

- [ ] Updated documentation


dremin (approved):
This is a great improvement, thanks!

PR #345, opened 2023-09-15
Reimplement chat-to-video around video demo app

dremin:
### Summary

This PR removes the `video-app-quickstart` app and replaces it with @mschmitt19's [twilio-video-demo-app](https://github.com/mschmitt19/twilio-video-demo-app), which is a more robust and full-featured interface for video rooms.

As a result of this change, several other changes have been made to simplify the solution:
- Flex now iframes the video app instead of having separate interfaces for agent and customer. As a result, the Flex plugin no longer has any video code, and has removed the alarmingly large `twilio-video` dependency. This should reduce our bundle size noticeably.
- Previously, a Sync map was used to correlate the task to a video room, and to provide automatic expiration. Now, Sync has been completely removed from the solution, in favor of putting the video room in a task attribute and using video's built-in room idle expiration.
- Refactored the Flex plugin code to use an `ApiService` for hitting serverless functions. This simplifies component logic.

The video demo app has also had several modifications to improve the user experience with the chat-to-video use case.

### Checklist

- [x] Updated documentation


jhunter-twilio (approved):
this is awesome, fantastic job!

lhamoudi:
Incredible! The emojis not showing was a known issue too before. Excited to play with this!

PR #347, opened 2023-09-18
Replace ticker with Redux for pending activity

dremin:
### Summary

Previously, the pending activity component would run a ticker to check the pending activity from local storage every second. This was quite inefficient (and could lag up to one second after a change); now instead we use Redux and `useSelector`.

### Checklist


jhunter-twilio (approved):
nice bit of optimizing!

PR #348, opened 2023-09-18
Enhancements to caller-id feature

dremin:
### Summary

1. Previously the list included only purchased Twilio phone numbers. Added new option (`include_outgoing_only_numbers`, enabled by default) to also fetch and include verified outgoing caller IDs to the list.
2. Previously the default value was the first item in the list. Now it uses the configured Flex dialpad default.
3. Previously if the selected caller ID was no longer a valid caller ID, the dropdown selection (which defaulted to the first item) and the configured value (now invalid) would get out of sync, causing failed dials and confusion. Now the default is used in this scenario.

### Checklist

- [x] Updated documentation

### DIFF
--- a/plugin-flex-ts-template-v2/src/utils/serverless/PhoneNumbers/PhoneNumberService.ts
+++ b/plugin-flex-ts-template-v2/src/utils/serverless/PhoneNumbers/PhoneNumberService.ts
@@ -41,13 +41,19 @@ class PhoneNumberService extends ApiService {
 
   private STORAGE_KEY_FLEX_PHONE_NUMBERS = `FLEX_PHONE_NUMBERS_${this.flex_service_instance_sid}`;
 
+  private STORAGE_KEY_FLEX_PHONE_NUMBERS_WITH_OUTGOING = `FLEX_PHONE_NUMBERS_WITH_OUTGOING_${this.flex_service_instance_sid}`;
+
   private STORAGE_KEY_VALIDATED_NUMBERS = `FLEX_VALIDATED_NUMBERS`;
 
   private EXPIRY = 86400000; // 1 day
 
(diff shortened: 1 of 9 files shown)


dremin:
@jhunter-twilio If you think the outgoing verified numbers should not be shown by default, I can do that. I don't feel strongly either way.

plugin-flex-ts-template-v2/src/utils/serverless/PhoneNumbers/PhoneNumberService.ts
```diff
@@ -45,13 +45,13 @@ class PhoneNumberService extends ApiService {
 
   private EXPIRY = 86400000; // 1 day
 
-  async getAccountPhoneNumbers(): Promise<ListPhoneNumbersResponse> {
+  async getAccountPhoneNumbers(includeOutgoing: boolean): Promise<ListPhoneNumbersResponse> {
```
jhunter-twilio, line 48: do we need something to distinguish here whether the cached result includes outgoing or not?  if you have the flag set to false then disable it and vice versa, it should renew the cache.  Perhaps the easiest way to do it is to have two separate cache keys

jhunter-twilio (changes requested):

plugin-flex-ts-template-v2/src/utils/serverless/PhoneNumbers/PhoneNumberService.ts
dremin: Agreed, done!

jhunter-twilio:
> @jhunter-twilio If you think the outgoing verified numbers should not be shown by default, I can do that. I don't feel strongly either way.

yeah im indifferent, default on is fine.

jhunter-twilio (approved):
LGTM 🥳

PR #358, opened 2023-09-25
DEMOENG-1270 - Restructure template documentation

dremin:
### Summary

Complete overhaul of the documentation structure. Much of the existing content has been retained within the new structure, and there is a fair amount of new content as well.

You can view the updates locally by running `npm run start:docs`.

I believe there are still more opportunities for edits to the existing content, but I think this is already a large improvement!

### Checklist

- [x] Tested changes end to end
- [x] Updated documentation
- [x] Requested one or more reviewers

PR #360, opened 2023-09-26
Add option to overwrite environment config

dremin:
### Summary

By default, the setup script will not erase an existing file; it will only update it if it exists. However, this means that if additional env vars have been later added in the example file or if one needs to be changed, re-running the setup script is not sufficient to get the changes added to the existing env file.

This PR adds an `--overwrite` option to the setup script allowing it to replace existing files with the updated example/template file.

Because the docs for this are being redone, I will document it in #358.

### Checklist

- [x] Updated documentation


jhunter-twilio (approved):
this looks great, the overwrite value is not in use yet anywhere though right? would we want to add it to the postinstall command for the top level package.json?

dremin:
Correct, it's not used other than being present in documentation. I didn't apply it to the postinstall command; we can do that in a separate PR if you think it makes sense.

PR #362, opened 2023-09-28
Custom Logging Solution

johnfischelli:
### Summary

Re-wrote the logging implementation to be driven by a custom class rather than a library.

We can now register logging hooks in a fashion similar to other hooks in the template. The Feature Loader is now aware of `loggerHooks`.

These Logger Hooks are actually "Destinations" for the logs themselves. The Logging class by default is not configured with any destinations, so the logs would essentially be forgotten. In addition to enhancing the template itself and its ability to load logging hooks - this includes a Feature that writes any logs to the console.

**Anyone is now free to write additional features that can send logs anywhere you like!**

### Destinations

Destinations are classes that implement sending logs to a specific location. Destinations can be configured with a log level. It will ignore lower level logs than its configured minimum log level.

This means you could register a destination that was only interested in receiving errors, but not debug logs, etc.

Destinations extend an abstract class provided by this logging enhancement. Here is an example destination that sends all logs to the console! This is the actual implementation used by the feature added in this PR.

```
import { LogLevel } from '../../../utils/logger';
import Destination from '../../../utils/logger/destination';

export default class Console extends Destination {
  async handle(level: LogLevel, message: string, context?: object, meta?: object): Promise<void> {
    return new Promise((resolve) => {
      let additonalContext: object = {};
      additonalContext = Object.assign(additonalContext, meta, context);

      // if there is additional context, use it
      if (Object.keys(additonalContext).length) {
        console[level](message, additonalContext);
      } else {
        console[level](message);
      }

      return resolve();
    });
  }
}
```

### Yes, you can still use console.log if you want
Developers using the template are still free to use `console.log` directly in their features - but I would recommend that after development is complete they switch to the logging library with appropriate log levels so that customers can configure where logs might go and what levels they are interested in.

### Metadata and Context 
In addition, our logger does support both metadata and context objects with its log messages. Metadata will be automatically appended to each log message. By default, we include the current worker sid and worker name to each log message when you use the logging library - but other forms of meta data would be possible.

Additionally, each Destination can be configured with its own metadata - or told to ignore all metadata entirely. For example, the console destination now ignores global metadata by default.

### How to use

In most cases, migrating to this logger is easy. Consider the log:

```console.log(`activity-reservation-handler: handle ${eventName} for ${task.sid}`);```

Can now be written as: 

```
import logger from 'utils/logger';
logger.log(`activity-reservation-handler: handle ${eventName} for ${task.sid}`);
```

The logger has methods for different log levels

```
import logger from 'utils/logger';

logger.debug('...');
logger.log('...');
logger.info('...');
logger.warn('...');
logger.error('...');
```
#### Adding Context
This logger supports adding additional context with your log message (string).

```
import logger from 'utils/logger';

logger.debug('This is my message', {task, worker});
```

### Log Buffering
Because the feature-loader takes time and loads features in alphabetical order, its possible that features are writing logs before all the destinations have been loaded from all other features. In order to ensure that these messages are not missed, the logger now buffers messages while there are no destinations configured. Upon completion of the feature-loader process the log will attempt to flush these buffered messages to all configured destinations.

### Checklist

- [ ] Updated documentation
- [ ] Requested one or more reviewers

### CODE REVIEW

File: plugin-flex-ts-template-v2/src/utils/feature-loader/logger.ts [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+    logger.addMetaData('workerName', worker.name);
+    logger.addMetaData('source', 'twilio:flex-ui');
+  }
+  logger.debug('logger initialized!');
+};
+
+export const processHooks = () => {
+  destinations.forEach((dest) => logger.addDestination(dest));
+  logger.processBuffer();
+};
```
dremin, line 20: I think these functions can be combined into a single `init` function, as the metadata is not relevant until the buffer is flushed. Then, the feature loader can run the `init` function as part of the `execute deferred hooks` section of code. This makes things more consistent with the other hooks.

plugin-flex-ts-template-v2/src/utils/feature-loader/teams-filters.ts
```diff
@@ -34,6 +36,6 @@ export const init = async (flex: typeof Flex, manager: Flex.Manager) => {
 };
 
 export const addHook = (flex: typeof Flex, manager: Flex.Manager, feature: string, hook: any) => {
-  console.info(`Feature ${feature} registered teams filter hook: %c${hook.teamsFilterHook.name}`, 'font-weight:bold');
+  logger.info(`Feature ${feature} registered teams filter hook: %c${hook.teamsFilterHook.name}`);
```
dremin, line 39: This log has a `%c` for formatting. We need to either remove these or find a way to format console logs that pass through the logger.

plugin-flex-ts-template-v2/src/feature-library/console-log-integration/config.ts
```diff
@@ -0,0 +1,13 @@
+import { getFeatureFlags } from '../../utils/configuration';
+import ConsoleLogIntegrationConfig from './types/ServiceConfiguration';
+
+const { enabled = false, log_level = 'info' } =
```
dremin, line 4: This is probably the only feature we actually want to default `enabled` to `true` here. This definition is primarily useful for when config is missing, and logging will be essential in such a situation.

dremin (changes requested):
This is very nice! A couple small items to take a look at.

plugin-flex-ts-template-v2/src/utils/feature-loader/teams-filters.ts
dremin: Let's keep these feature loader logs using `console.info`. It gives us the formatting, plus if there is an issue with bootstrapping then we would want the logs always going to the console anyway.

plugin-flex-ts-template-v2/src/utils/feature-loader/logger.ts
johnfischelli: Done! We now call init once and dont have to call any additional methods like other deferred hooks!

plugin-flex-ts-template-v2/src/utils/feature-loader/teams-filters.ts
johnfischelli: Done!

plugin-flex-ts-template-v2/src/feature-library/console-log-integration/config.ts
johnfischelli: Done!

plugin-flex-ts-template-v2/src/utils/feature-loader/logger.ts
```diff
@@ -24,5 +20,6 @@ export const addHook = (feature: string, hook: any) => {
     logger.warn(`Feature ${feature} declared logger hook, but is missing loggerHook to hook`);
     return;
   }
+  console.info(`Feature ${feature} registered logger hook: %c${typeof hook.loggerHook}`, 'font-weight:bold');
```
dremin, line 23: I'd like to go for consistency here and log the function name if possible! You'll need to adjust the hook to have a function name as well.

dremin (approved):
Looks great! 🎉 Will merge once the docs are in.

jhunter-twilio (commented):
This is an excellent addition, I love it.

The only thing I can think of, do we need a special consideration for the remove-feature script though when removing this feature?

dremin:
> This is an excellent addition, I love it.
> 
> The only thing I can think of, do we need a special consideration for the remove-feature script though when removing this feature?

That's a fair point. While it wouldn't be catastrophic, it would prevent the remaining feature's logs from working. I see a couple ways around this:
- Exclude the feature from the remove-features script
- Bundle the console logger with the template as a util rather than as a feature
- Do nothing from the code perspective, and document "features you might want to keep" in the remove-features script doc.

jhunter-twilio:
Yeah I think the best option is to consider this as part of the template rather than a traditional feature. In that sense the documentation should align with that.

I think we still need to update the remove features script to actively ignore it then.

PR #363, opened 2023-09-28
Further enhance CRM container with extensible tabs

dremin:
### Summary

- Refactors the feature to be an extensible tabs container
- Features can register tabs using an actions hook
- The enhanced-crm-container feature registers a tab for the iframe
- Maintains the existing functionality to persist the component across task switches, as well as parentTask logic
- Simplified 'no tasks' logic; it is the decision of the feature registering a tab whether or not to register based on the payload task.
- Refactored canned-responses CRM mode to instead register a tab if the enhanced-crm-container feature is enabled.

Future things to improve upon:
- Ship a simple webpage asset in place of Bing
- Find a way to scroll the tab panel instead of the tabs scrolling away

### Checklist

- [x] Updated documentation


plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/flex-hooks/events/pluginsInitialized.ts
jhunter-twilio, line 1: this is a little bit of a departure from how we register custom actions - just for discoverability/readability would you mind moving the register action to a separate file? i realize the file is largely going to be empty as the action doesn't do anything but it took me a few minutes to find this because i was expecting to see a custom-action folder haha

plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/flex-hooks/events/pluginsInitialized.ts
```diff
@@ -0,0 +1,10 @@
+import * as Flex from '@twilio/flex-ui';
+
+import { FlexEvent } from '../../../../types/feature-loader';
+
+export const eventName = FlexEvent.pluginsInitialized;
+export const eventHook = function registerAction(flex: typeof Flex, _manager: Flex.Manager) {
+  flex.Actions.registerAction('CRMContainerLoaded', async () => {
```
jhunter-twilio, line 7: i realize why you've called it CRMContainerLoaded but I wonder if something like LoadCRMContainerTabs would be more appropriate and intuitive since the way we are using it (if im following correctly) is to use the before hook to capture them all when invoking the event

jhunter-twilio (changes requested):
I've just taken a quick first pass through this, i've yet to run it and test it but i wanted to get these thoughts out sooner rather than later - let me know what you think.

plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/flex-hooks/events/pluginsInitialized.ts
jhunter-twilio: maybe thats not the best name either, but something thats a little more intuitive

plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/flex-hooks/events/pluginsInitialized.ts
dremin: It looks like a few features have it like this, I'll fix them all!

plugin-flex-ts-template-v2/src/feature-library/enhanced-crm-container/flex-hooks/events/pluginsInitialized.ts
dremin: Ha, I struggled with this name too 😄 I am good with `LoadCRMContainerTabs`!

jhunter-twilio (approved):
this is excellent, sorry for the slow review!

PR #365, opened 2023-09-28
Pre-configure API key/secret and local flex-config for easier deploy

dremin:
### Summary

Currently the steps to do a deploy of flex-config from the developer machine are harder than necessary. The setup script should be able to do all of the setup tasks. This PR makes the necessary updates such that the deploy command works OOB. As a bonus, it makes local video work OOB as well!

- When running the setup script locally, it will create the `ui_attributes.local.json` file and fill the vars
- Updated the default environment value to `local`
- Updated the setup script to fill API key and secret from the local CLI profile config JSON

Also performed some updates to the relevant docs sections:
- Revamped and moved local flex-deploy steps to the local deploy doc
- Fixed commands to check the account SID
- Moved the taskrouter skills section from package overview to the configuration doc

### Checklist

- [x] Updated documentation


dremin:
Confirmed this also works on Windows! 🪟

PR #367, opened 2023-09-28
Add new asset to use as default CRM IFrame page

dremin:
### Summary

Adds a basic HTML asset to the serverless deployment that displays the contents of two query params, and updated the default `enhanced-crm-container` URLs to use the new asset instead of Bing.

Enhanced the `enhanced-crm-container` feature to add a template string replacement for the serverless URL and updated docs accordingly.

### Checklist

- [x] Updated documentation


jhunter-twilio (approved):
i like it 💯

PR #369, opened 2023-10-02
Updates for Flex Plugins CLI 6.2.0

dremin:
### Summary
6.2.0 was released on 9/28 and brought a breaking change along with it--an upgrade to Webpack 5.

One of our dependencies references the `global` object, which is no longer polyfilled by default in Webpack 5. I have re-enabled this in the webpack config.

Tested that this change does not regress existing users on 6.1.1. Also validated that the webpack change did not change the plugin size. Also validated deployment still works correctly.

### Checklist


jhunter-twilio (approved):
Lgtm 👍

PR #376, opened 2023-10-04
Common configuration for sharing across features

dremin:
### Summary

We've been running into more situations where config that is specifically per-feature is causing problems. For example, the list of teams should be shared across the various teams view features.

Posting this as a draft to gather feedback on the approach and to ensure this covers all of the use cases we want it to.

### Checklist

- [x] Updated documentation

### DIFF
--- a/docs/docs/building/template-utilities/configuration.md
+++ b/docs/docs/building/template-utilities/configuration.md
@@ -39,6 +39,9 @@ The custom_data model that lives in ui_attributes follows this schema:
 ```json
 {
 "custom_data": {
+    "common": {
+      // template-wide config options are here
+    },
     "features": {
       "<feature-name>": {
         "enabled": true,
@@ -54,9 +57,22 @@ The custom_data model that lives in ui_attributes follows this schema:
 
 Ultimately, enablement of each feature is managed by this object as it appears in the [hosted Flex configuration](https://www.twilio.com/docs/flex/developer/config/flex-configuration-rest-api#ui_attributes) (or `appConfig.js` if running [locally](#local-environment))
 
+#### Common configuration
+
+Some configuration is not specific to a feature, and is included in the `common` section. This section is intended to be used for items that could be used by several features.
+
+:::tip Food for thought
+If you have two features which need to share a piece of configuration, consider that it likely makes more sense to combine the features rather than using common configuration.
+:::
+
+The following common configuration properties are included by default:
+
+- **`log_level`** - The minimum log level to output to the browser console. `info` by default; may be set to `debug`, `log`, `warn`, `info`, or `error`
+- **`teams`** - Array of team names used by various features to populate team lists, matching the values used in worker attributes.
+
 #### Per-worker configuration overrides
 
-When performing testing, or for managing a diverse contact center, you may wish different workers to have a different feature configuration. This could be variations of feature settings or enabling a different feature set altogether. You may add a `config_overrides` object to a worker's attributes to override any feature configuration, taking priority over both the hosted Flex configration and the local `appConfig.js` (if applicable).
+When performing testing, or for managing a diverse contact center, you may wish different workers to have a different feature configuration. This could be variations of feature settings or enabling a different feature set altogether. You may add a `config_overrides` object to a worker's attributes to override any feature configuration, taking priority over both the hosted Flex configuration and the local `appConfig.js` (if applicable).
 
 For example, if the `activity-reservation-handler` feature is globally enabled but you wish to disable it for a specific worker, you can add the following to the worker's attributes to disable it:
 
--- a/docs/docs/building/template-utilities/configuration.md
+++ b/docs/docs/building/template-utilities/configuration.md
@@ -69,10 +69,11 @@ The following common configuration properties are included by default:
 
 - **`log_level`** - The minimum log level to output to the browser console. `info` by default; may be set to `debug`, `log`, `warn`, `info`, or `error`
 - **`teams`** - Array of team names used by various features to populate team lists, matching the values used in worker attributes.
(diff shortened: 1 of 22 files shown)


johnfischelli:
I really like this!

I ran this branch locally but forgot to update my configuration api and when I opened the common settings:
image

I have to assume this would be a common occurrence. Can we handle it more gracefully? Any way to still render the settings even if they haven't been added to the configuration api yet?

johnfischelli:
Another general thought - do these common settings need more explanation and context? Previously, when settings were contextualized to a feature I feel like they were more clear.

With common settings, there is more ambiguity. For example, the common log level. It will most likely only affect console logs, but that isn't immediately clear because there is no way to add the necessary description or context.

I think the simplest way to handle this for now would be to have docs for common settings.

ldvlgr:
Just a note to update the Feature docs.
https://twilio-professional-services.github.io/flex-project-template/feature-library/teams-view-enhancements#teams-view-data-tiles
Here I have this in the docs:
 _This feature uses the list of Team names as configured in the Teams View Filters feature and the Activity configuration (colors and icons) from the Queues View Data Tiles._

Should we also move the **agent activities** section to common? ( I use that in 2 features)

dremin:
> Another general thought - do these common settings need more explanation and context? Previously, when settings were contextualized to a feature I feel like they were more clear.
> 
> With common settings, there is more ambiguity. For example, the common log level. It will most likely only affect console logs, but that isn't immediately clear because there is no way to add the necessary description or context.
> 
> I think the simplest way to handle this for now would be to have docs for common settings.

This is also something that I was concerned about. This area could get messy, and it also wouldn't be affected by things like the remove-features script, so if you remove all the consumers of a setting it would still remain there. I wonder if we should try and limit the usage of this section as much as possible, and steer developers towards combining features rather than including a bunch of settings that are used by 2 features?

docs/docs/building/template-utilities/configuration.md
```diff
@@ -54,9 +57,22 @@ The custom_data model that lives in ui_attributes follows this schema:
...
+- **`teams`** - Array of team names used by various features to populate team lists, matching the values used in worker attributes.
+
 #### Per-worker configuration overrides
 
-When performing testing, or for managing a diverse contact center, you may wish different workers to have a different feature configuration. This could be variations of feature settings or enabling a different feature set altogether. You may add a `config_overrides` object to a worker's attributes to override any feature configuration, taking priority over both the hosted Flex configration and the local `appConfig.js` (if applicable).
+When performing testing, or for managing a diverse contact center, you may wish different workers to have a different feature configuration. This could be variations of feature settings or enabling a different feature set altogether. You may add a `config_overrides` object to a worker's attributes to override any feature configuration, taking priority over both the hosted Flex configuration and the local `appConfig.js` (if applicable).
```
jhunter-twilio, line 75: is it worth mentioning whether common config can be overridden at the worker level?

docs/docs/feature-library/teams-view-filters.md
```diff
@@ -66,17 +66,6 @@ To use the sample features, most of the filters can simply be toggled on in the
           "Purchasing",
```
jhunter-twilio, line 66: food for thought, granted department isnt actually required by any other features but i feel like it intuitively belongs in the common area of configuration now.  What do you think?

plugin-flex-ts-template-v2/src/feature-library/admin-ui/utils/helpers.ts
```diff
@@ -7,6 +7,11 @@ import AdminUiService from './AdminUiService';
 
 const acronyms = ['id', 'ui', 'sip', 'pstn', 'sms', 'crm', 'sla', 'cbm', 'url'];
 const hiddenFeatures = ['admin_ui'];
+const docsBaseUrl = 'https://twilio-professional-services.github.io/flex-project-template';
```
jhunter-twilio, line 10: nothing i think we need to address right now but later down the road we might want to think about doing something here that might dynamically link this to the repo it belongs to rather than the base template docs along with all the other admin panel docs links

plugin-flex-ts-template-v2/src/feature-library/teams-view-enhancements/config.ts
```diff
@@ -66,18 +68,6 @@ export const getEnabledChannels = () => {
   });
   return enabledChannels;
 };
-export const getChannelVoice_Color = () => {
```
jhunter-twilio, line 69: could you elaborate a little on why these got removed, are they unused?

jhunter-twilio (approved):
This looks great to me, ive added some comments if you want to give some feedback on them before merging but ultimately a thumbs up! 👍

ldvlgr (approved):
LGTM2!

plugin-flex-ts-template-v2/src/feature-library/admin-ui/utils/helpers.ts
dremin: Maybe it could be configurable, not sure how we could automatically deduce this reliably.

plugin-flex-ts-template-v2/src/feature-library/admin-ui/flex-hooks/strings/index.ts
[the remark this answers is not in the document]
dremin: Yes, but it needs to be translated by someone 😅

docs/docs/building/template-utilities/configuration.md
dremin: Sure is!

docs/docs/feature-library/teams-view-filters.md
dremin: I don't disagree! I will move this.

PR #383, opened 2023-10-11
correcting missing comma

jhunter-twilio:
### Summary

The enhanced crm container modifications missed a comma when updating the ui_attributes which causes installs to fail.


dremin (approved):
LGTM

PR #387, opened 2023-10-13
Package-based dynamic add-on deployments

dremin:
### Summary

POC of an alternative approach for dynamically deploying services, using a package-based approach rather than Terraform.

Advantages:
- Much simpler
- Not limited to Twilio serverless projects--can be used for any npm package that needs its deployment orchestrated with the template
- When working with serverless projects, provides a consistent developer experience working with the serverless toolkit to test services locally, as well as taking advantage of existing deployment logic within the toolkit

Disadvantages:
- Does not provide an example of Terraform serverless deployments
- No mechanism for resolving cross-dependencies between packages
  - A solution to this could be the 'initial deploy' process we already have

### Checklist

- [x] Updated documentation


jhunter-twilio:
I like the simplicity of this for sure - @lvaladares, what do you think - does this address your needs?

lvaladares (approved):
yes @jhunter-twilio
cleanly an simply
lgtm

scripts/remove-features.mjs
```diff
@@ -164,16 +156,16 @@ const performRemovals = async () => {
   
   if (!keepFeatures.includes("schedule-manager")) {
```
jhunter-twilio, line 157: this area could get tricky to maintain in the future as we add in serverless packages but we can probably refactor this slightly later.

jhunter-twilio (approved):
this is looking great

dremin:
Re-requesting review due to the docs changes.

jhunter-twilio (approved):
yeah i like these doc updates too - its starting to look pretty sensible now :) once we get the terraform changes in we can round it all off with documentation there too.

PR #420, opened 2023-11-01
Worker activity icons

ldvlgr:
### Summary

Adding Worker Activity Icons for each worker on the Teams View. The UserCard component is not customizable. This adds a larger colored circle with icon to show the agent's activity.  Activities Config is re-used from the Metrics Tiles.

Also included here is a new Agent Activity column for the Queues View that shows a Data Icon. Clicking on the data icon will show a small Paste Popover window with a breakdown of all agents by activity, for example:

**(Queue name)
Agent Activity Summary**
Available: 3
Break: 2
Training: 1
Offline: 4

### Checklist

- [x] Updated documentation


plugin-flex-ts-template-v2/src/feature-library/metrics-data-tiles/custom-components/AgentActivityTile/AgentActivityTile.tsx
```diff
@@ -57,23 +58,25 @@ const AgentActivityTile = (props: ComponentProps) => {
...
-          );
-        })}
-        <AgentActivity key="other">
-          <Icon icon={activityConfig.other?.icon} />
-          <Label bgColor={activityConfig.other?.color}>
-            <Template source={templates[StringTemplates.Other]} />
-          </Label>
-          <Metric> {activityCounts.other} </Metric>
-        </AgentActivity>
+        <Box overflowY="auto" maxHeight="180px">
```
dremin, line 61: 180px causes clipping for me with the defaults. 190px seems to be better.

plugin-flex-ts-template-v2/src/feature-library/teams-view-enhancements/config.ts
```diff
@@ -10,17 +10,28 @@ const {
 } = (getFeatureFlags()?.features?.teams_view_enhancements as TeamsViewEnhancementsConfig) || {};
 
 const {
+  calls = true,
+  other_tasks = true,
   team = true,
   department = false,
   location = false,
   agent_skills = true,
+  activity_icon = true,
```
dremin, line 19: This goes with the earlier comment to disable the icon column by default.

plugin-flex-ts-template-v2/src/feature-library/teams-view-enhancements/flex-hooks/components/AgentActivityIcon/AgentActivityIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+      [key: string]: ActivityConfig;
+    };
+    other: ActivityConfig;
+  };
+}
+
+const AgentActivityIcon = (props: ComponentProps) => {
+  const { activityName, activityConfig } = props;
+  let bgColor = activityConfig.other.color;
+  let icon = activityConfig.other.icon;
```
dremin, line 25: I think we need to do some checks here if activityConfig is incomplete or missing. If someone removed the data tiles feature, it would be missing, so some sort of fallback should be in place.

dremin (changes requested):

plugin-flex-ts-template-v2/src/feature-library/teams-view-enhancements/flex-hooks/components/AgentActivityIcon/AgentActivityIcon.tsx
dremin: Another option rather than doing it here, would be to do the null checks within config.ts.

ldvlgr: so in case "metrics_data_tiles" can not be found in the ui_attributes, it will use that configuration

flex-config/ui_attributes.common.json
```diff
@@ -251,10 +251,13 @@
...
         "display_task_queue_name": true,
         "columns": {
+          "calls": true,
+          "other_tasks": true,
           "team": true,
           "department": false,
           "location": false,
-          "agent_skills": true
+          "agent_skills": true,
+          "activity_icon": true
```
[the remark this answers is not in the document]
ldvlgr: done

plugin-flex-ts-template-v2/src/feature-library/teams-view-enhancements/config.ts
ldvlgr: done

dremin (approved):
🎉

PR #312, opened 2023-08-07
Updating activity skill filter

jhunter-twilio:
### Summary

updated activities skill filter to 

- allow for rules by skill name (backwards compatible with skill sid)
- fixed the ordering feature which broke in flex 2.x
- enhanced the ordering to use sort_order then activity name
- no longer require a rule for all activities, if a rule is missing it is assumed to be visible and alphabetically sorted at the start of the list (sort order 0)

### Checklist

- [x] Tested changes end to end
- [x] Any config changes added to all environment files
- [x] Completed README template for new features in `/docs/docs/feature-library/` and added feature row entry into the `/docs/docs/feature-library/00_overview.md` file
- [x] Added PR Label "ready-for-review"
- [x] Requested one or more reviewers for the PR

PR #426, opened 2023-11-06
Enhanced wrap up form

ldvlgr:
### Summary

Enhancing the Dispositions Wrap-up form with the ability to configure additional Text attributes or Select type attributes (pick from a list). Each custom attribute has a `form_label` and mapping to a `conversation_attribute` (for Flex Insights reporting). 

### Checklist

- [x] Updated documentation


plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -114,6 +138,49 @@ const DispositionTab = (props: OwnProps) => {
             ))}
           </RadioGroup>
         )}
+        {selectAttributes.map((attr) => {
+          const id = attr.conversation_attribute;
+          return (
+            <>
+              <Label htmlFor={id} required>
```
dremin, line 145: These don't actually behave as `required`. Either the required flag should be removed or the logic in `CompleteTask` should be updated to validate the custom fields (I suppose it might also make sense to add settings per attribute as to whether each is required).

dremin (changes requested):

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
ldvlgr: Wrap Up form attributes can now optionally be marked as `required: true` via the config
CompleteTask will pop a notification when required fields are missing:
_"Please select an option or enter a value for all required fields"_

flex-config/ui_attributes.common.json
```diff
@@ -224,7 +224,10 @@
             "require_disposition": true,
             "dispositions": ["Promotional Sale", "Renewal"]
           }
-        }
+        },
+        "text_attributes": [],
+        "select_attributes": [],
+        "multi_select_group" : {}
```
jhunter-twilio, line 230: this config should probably be supported by queue like the disposition config.  

I think we also want to change the "queue" configuration to support queue name as well as queue sid - i done something similar for [activity skill filter](https://github.com/twilio-professional-services/flex-project-template/pull/312/files) that might be worth taking a look at. Could i ask you make the tweak to match config on queue name or queue sid since you are in this space?

jhunter-twilio (changes requested):
Hey @ldvlgr 

Its looking really good - i can't wait to add it in - just the two requests to the way config is matched

1. config by queue
2. queue matched by name as well as queue sid

jhunter-twilio (commented):
As i think about this more - the config seems inconsistent with the way its laid out by object - but if we change it, its a breaking change - on the other hand i'd prefer to get the config sensibly organized now rather than leaving it looking inconsistent and frustrating to use

current model

```
{
        "global_dispositions": [],
        "per_queue": {
          "queuesid": {
            "require_disposition": true,
            "dispositions": [""]
          }
        },
        "text_attributes": [],
        "select_attributes": [],
        "multi_select_group" : {}
}
```

i think ideally we would want something like

```
{
  "global": {
    "dispositions" : [],
    "text_attributes" : [],
    "select_attributes": [],
    "multi_select_group" : {}
  },
  "per_queue": {
    "queuesid": {
      "dispositions" : [],
      "text_attributes" : [],
      "select_attributes": [],
      "multi_select_group" : {}
    }
  },
}
```

I think it's better we organize the config the right way now and impose the breaking change - consumers of updates will need to merge their config into the new model, i don't think it should be too complicated and fairly self evident for anyone doing a merge.

Whats your thoughts @dremin @mvrzan @ldvlgr ?

flex-config/ui_attributes.common.json
ldvlgr: I think we may be over-engineering this and adding unnecessary levels of complexity that most customers probably won't need.  Typically a wrap-up form contains a set of generic questions that apply to all types of calls/contacts. Any Text and Select fields on this wrap-up form can already be marked as NOT required so it's flexible enough to support different use cases where not all options may not be applicable for each call/queue.  For example, a list of Topics (Reasons for the Call) should be a comprehensive superset of all possible reasons customer could be calling/contacting a business.  We should try to balance Perfection vs Good Enough. I think what we provide here in the Extended Wrap up form with customizable Text Boxes and Pick from a list (Select/Pulldown menu) should cover the vast majority of use cases.

ldvlgr: Done.  You can now also configure all attribute types (Text, Select, Checkboxes) per queue.  Updated to code to match by queue name.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -50,23 +72,45 @@ const DispositionTab = (props: OwnProps) => {
...
     }
-
+    payload.custom_attributes = { ...customAttributes };
     dispatch(updateDisposition(payload));
   };
 
   const updateStoreDebounced = debounce(updateStore, 250, { maxWait: 1000 });
 
   useEffect(() => {
+    const getQueueFromInstantQuery = async (queueSid: string) => {
```
jhunter-twilio, line 82: Are we doing this look up just to get the queue name? If so can we remove it, the queue name is already on the task task.queueName, task.queueSid

plugin-flex-ts-template-v2/src/feature-library/dispositions/flex-hooks/actions/CompleteTask.ts
```diff
@@ -26,19 +36,23 @@ export const actionHook = function setDispositionBeforeCompleteTask(flex: typeof
     if (!payload.task?.taskSid) {
       return;
     }
-
-    const numDispositions = getDispositionsForQueue(payload.task?.queueSid ?? '').length;
+    let queueName = '';
+    const queue = await FlexHelper.getQueue(queueName);
+    if (queue) queueName = queue?.queue_name;
+    const numDispositions = getDispositionsForQueue(queueName ?? '').length;
```
jhunter-twilio, line 42: can we modify these functions to support looking up by queue name and queue sid - if we don't also look up by queue sid - it will break any existing configurations

jhunter-twilio (changes requested):
Almost there - just a couple tidy up items!

plugin-flex-ts-template-v2/src/feature-library/dispositions/flex-hooks/actions/CompleteTask.ts
ldvlgr: Do you expect customer to deploy the updated Template plugin code but NOT the updated config?

jhunter-twilio: No but they will have to merge it

ldvlgr: Ok, done!

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -78,6 +122,10 @@ const DispositionTab = (props: OwnProps) => {
     updateStoreDebounced();
   }, [notes]);
 
+  useEffect(() => {
```
mvrzan, line 120: I believe you can combine the above `useEffect` with this one. Just have both `customAttributes` and `notes` as a `useEffect` dependency.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -50,23 +72,45 @@ const DispositionTab = (props: OwnProps) => {
...
       }
 
-      if (isNotesEnabled() && tasks[props.task.taskSid].notes) {
-        setNotes(tasks[props.task.taskSid].notes);
+      if (isNotesEnabled()) {
+        setNotes(tasks[props.task.taskSid].notes || '');
       }
+      // set custom attributes from Redux state
+      setCustomAttributes(tasks[props.task.taskSid].custom_attributes);
+      const options = {} as any;
```
mvrzan, line 93: Would be great if you can define the type of the object rather than using `any` type.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -50,23 +72,45 @@ const DispositionTab = (props: OwnProps) => {
...
+      const optionsString = tasks[props.task.taskSid].custom_attributes[group.conversation_attribute] || '';
+      optionsString.split('|').forEach((opt) => {
+        options[opt] = true;
+      });
+      setGroupOptions(options);
+
+      const optionsForQueue = {} as any;
```
mvrzan, line 100: Would be great if you can define the type of the object rather than using any type.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -50,23 +72,45 @@ const DispositionTab = (props: OwnProps) => {
     if (isNotesEnabled() && notes) {
       payload.notes = notes;
     }
-
+    payload.custom_attributes = { ...customAttributes };
     dispatch(updateDisposition(payload));
   };
 
   const updateStoreDebounced = debounce(updateStore, 250, { maxWait: 1000 });
```
mvrzan, line 79: This line is only used within the `useEffect`. You could move this directly into the `useEffect`.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -50,23 +72,45 @@ const DispositionTab = (props: OwnProps) => {
...
 
   useEffect(() => {
```
mvrzan, line 81: This `useEffect` has a lot of things going on. Here's a suggestion that might make this component a bit smaller, and easier to read. You could pull this into a custom react hook and just invoke it here. That way this whole `useEffect` block could be replaced with a single line.

mvrzan (changes requested):

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -114,6 +192,107 @@ const DispositionTab = (props: OwnProps) => {
...
+          <div key={groupForQueue.conversation_attribute}>
+            <CheckboxGroup
+              orientation="horizontal"
+              name={groupForQueue.conversation_attribute}
+              legend={groupForQueue.form_label}
+              helpText={templates[StringTemplates.ChooseOptions]()}
+              required={groupForQueue.required || false}
+            >
+              {groupForQueue.options.map((option) => {
+                return (
```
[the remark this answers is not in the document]
ldvlgr: Done!

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
ldvlgr: Done!

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -4,25 +4,46 @@ import { Actions, ITask, Template, templates, withTaskContext } from '@twilio/fl
...
 export interface OwnProps {
   task?: ITask;
 }
 
 const DispositionTab = (props: OwnProps) => {
   const NOTES_MAX_LENGTH = 100;
+  const [queueName, setQueueName] = useState('');
   const [disposition, setDisposition] = useState('');
   const [notes, setNotes] = useState('');
+  const [customAttributes, setCustomAttributes] = useState({} as any);
```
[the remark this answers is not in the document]
ldvlgr: added type

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -4,25 +4,46 @@ import { Actions, ITask, Template, templates, withTaskContext } from '@twilio/fl
...
+  const [groupOptions, setGroupOptions] = useState({} as any);
+  const [groupOptionsForQueue, setGroupOptionsForQueue] = useState({} as any);
```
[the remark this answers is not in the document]
ldvlgr: add types

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -114,6 +192,107 @@ const DispositionTab = (props: OwnProps) => {
...
+          <div key={group.conversation_attribute}>
+            <CheckboxGroup
+              orientation="horizontal"
+              name={group.conversation_attribute}
+              legend={group.form_label}
+              helpText={templates[StringTemplates.ChooseOptions]()}
+              required={group.required || false}
+            >
+              {group.options.map((option) => {
+                return (
```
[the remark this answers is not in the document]
ldvlgr: done

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -114,6 +192,107 @@ const DispositionTab = (props: OwnProps) => {
...
+                    name={option}
+                  >
+                    {option}
+                  </Checkbox>
+                );
+              })}
+            </CheckboxGroup>
+          </div>
+        )}
+        {selectAttributes.map((attr) => {
```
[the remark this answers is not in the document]
ldvlgr: perfect.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -34,6 +55,7 @@ const DispositionTab = (props: OwnProps) => {
       taskSid: props.task.taskSid,
       disposition: '',
       notes: '',
+      custom_attributes: {},
     };
```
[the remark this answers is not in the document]
ldvlgr: I left this for now. Store was not properly updating custom_attributes after simplifications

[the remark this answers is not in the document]
ldvlgr: Updated and tested fine!  Much simpler now, thanks.

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/DispositionTab/DispositionTab.tsx
```diff
@@ -19,101 +30,235 @@ export interface OwnProps {
...
-    }
-
-    if (isNotesEnabled() && notes) {
-      payload.notes = notes;
-    }
-
     dispatch(updateDisposition(payload));
   };
 
   const updateStoreDebounced = debounce(updateStore, 250, { maxWait: 1000 });
```
mvrzan, line 72: I faced a similar problem in a different code base and I realized the `debounce` here is not used correctly. You are not debouncing anything here. You are just delaying the execution of `updateStore` for `250ms`. So if you type 10 characters in that time, at the end, you will send 10 requests to the `updateStore` function.

mvrzan (changes requested):

ldvlgr:
I think this is ready guys. Please review once more!  Thanks

plugin-flex-ts-template-v2/src/feature-library/dispositions/custom-components/Notes/Notes.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ITask, Template, templates } from '@twilio/flex-ui';
+import { TextArea } from '@twilio-paste/core/textarea';
+import { Label } from '@twilio-paste/core/label';
+import { HelpText } from '@twilio-paste/core/help-text';
+
+import { StringTemplates } from '../../flex-hooks/strings';
+
+export interface OwnProps {
+  task?: ITask;
+  notes: string;
```
dremin, line 11: This value doesn't seem to be used. I noticed a regression where notes are no longer saved/restored when switching views, and this may be part of the reason.

plugin-flex-ts-template-v2/src/feature-library/dispositions/flex-hooks/actions/CompleteTask.ts
dremin, line 1: This doesn't seem to handle checking if the `multi_select_group` is required or present, so that setting does not seem to do anything.

dremin (changes requested):

dremin (approved):
Better late than never! :tada:

PR #430, opened 2023-11-09
Refactor custom directory components for reusability, override worker directory

dremin:
### Summary

Now if someone wants to add their own tab, it is much easier:
- There is a common DirectoryTab component that includes all of the layout and search logic.
- There is a common DirectoryItem component for rows in DirectoryTab
- Now to add a new tab, just pass an array of DirectoryEntry objects, and a transfer function, to DirectoryTab

Added a custom worker directory tab to complement the queue tab. Now the worker tab can filter unavailable agents, and has the correct icons per media type.

### Checklist


plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/DirectoryTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <SearchBox key="key-tab-search-box" onInputChange={onQueueSearchInputChange} inputRef={searchInputRef} />
+      <Flex key="external-tab-results" vertical element="TRANSFER_DIR_COMMON_ROWS_CONTAINER">
+        {props.isLoading && (
+          <Flex hAlignContent="center">
+            <Spinner decorative size="sizeIcon90" />
+          </Flex>
+        )}
+        {filteredDirectory.length === 0 && !props.isLoading ? (
+          <Alert variant="neutral">
+            <Template source={templates[StringTemplates.NoItemsFound]} />
```
jhunter-twilio, line 69: Would it be possible to parameterize this a little further so we can have tab specific messages when no items are found?  I think it is useful to qualify the likely reason no items are found, i can foresee people raising issues when queues are not available because no one else is logged in etc.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/types/ServiceConfiguration.ts
```diff
@@ -1,3 +1,5 @@
+import { ExternalDirectoryEntry } from './DirectoryEntry';
+
 export default interface CustomTransferDirectoryConfig {
   enabled: boolean;
```
jhunter-twilio, line 4: i think this is missing the new worker config definition

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/config.ts
```diff
@@ -1,7 +1,7 @@
 import { Manager } from '@twilio/flex-ui';
 
 import { getFeatureFlags } from '../../utils/configuration';
-import { ExternalDirectoryEntry } from './types/ServiceConfiguration';
+import { ExternalDirectoryEntry } from './types/DirectoryEntry';
 
 const { enabled = false, use_paste_search_icon = false } = getFeatureFlags()?.features?.custom_transfer_directory || {};
```
jhunter-twilio, line 6: this should declare the type from the ServiceConfiguration  (i realize this is a hangover from when i wrote it 😊 )

jhunter-twilio (approved):
Looks great, i love the refactor - some minor updates to tackle plus an optional item on the _no items_ label

I think we want to update the documentation too, to conform to the broader standard but we can do that as part of a separate effort.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/custom-components/DirectoryTab.tsx
dremin: Added optional prop to DirectoryTab allowing a custom no entries message. The generic message is used as a fallback if none is provided, as well as for no search results.

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/types/ServiceConfiguration.ts
dremin: Fixed!

plugin-flex-ts-template-v2/src/feature-library/custom-transfer-directory/config.ts
dremin: Fixed!

jhunter-twilio (approved):
Awesome work!

PR #433, opened 2023-11-13
Adding function to support create-task from studio

jhunter-twilio:
### Summary

Adding a convenience function for creating a task from studio.  This is particularly useful when creation of a task is required when invoking a studio flow from the REST API.

### Checklist

### DIFF
--- a/serverless-functions/src/functions/common/studio/taskrouter/create-task.js
+++ b/serverless-functions/src/functions/common/studio/taskrouter/create-task.js
@@ -1,0 +1,34 @@
+const { prepareStudioFunction, extractStandardResponse } = require(Runtime.getFunctions()[
+  'common/helpers/function-helper'
+].path);
+const TaskRouterOperations = require(Runtime.getFunctions()['common/twilio-wrappers/taskrouter'].path);
+
+const requiredParameters = [
+  { key: 'workflowSid', purpose: 'workflow to execute' },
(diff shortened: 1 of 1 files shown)


serverless-functions/src/functions/common/studio/taskrouter/create-task.js
```diff
@@ -0,0 +1,34 @@
...
+    const { workflowSid, taskChannel, attributes, priority: overriddenPriority, timeout: overriddenTimeout } = event;
+
+    console.log('attributes: ', attributes);
+    const result = await TaskRouterOperations.createTask({
+      context,
+      workflowSid,
+      taskChannel,
+      attributes: JSON.parse(attributes),
+      priority: overriddenPriority || 0,
+      timeout: overriddenTimeout || 86400,
```
dremin, line 23: The wrapper defines defaults so I don't think we need to duplicate that here.

jhunter-twilio:
Queue-stats-metadata is probably no longer applicable with the new API that's available - that function and the purpose behind it probably needs a little rethinking.

I agree that this function along with the neighboring common functions and their associated flex util services need documentation but I think this is probably the wrong place to start, I think we need to restructure the docs slightly to sensibly accommodate that.

Any reservations if I tackle that in a separate PR that we can collaborate on?

Issue #441, opened 2023-11-27
Add option to use Virtual Start Time for callback tasks


chaosloth:
Call back tasks in the queued callback and voicemail feature are created without regard to the parent tasks time in queue, instead the priority is used to deliver callbacks first to agents, in effect adjusting the task order so that callbacks are delivered first.  

The request is that the 'virtualStartTime' is set on the child (callback) task using the original start time of the parent task. In this was the task order is maintained.

https://www.twilio.com/docs/taskrouter/queue-ordering-last-first-out-lifo

PR #442, opened 2023-11-28
Use virtualStartTime for callback requests

dremin:
### Summary

Addresses #441 by adding an option to the `wait-experience` logic, which when enabled (it is by default), sets the `virtualStartTime` parameter to the value of the original task's `dateCreated`.

Also updated the Twilio library to a version that supports the parameter.

### Checklist

- [x] Updated documentation


jhunter-twilio (approved):
So glad we can now do this!

PR #445, opened 2023-12-01
Dual Channel Recording Race Condition Fix

johnfischelli:
### Summary

instead of watching for both worker and customer participants - we're only ever recording one or the other.

this change will use the configured "channel to record" to narrow down to only a single participant

this prevents an issue - especially when recording the agent leg - that the flex state doesn't get the customer conference participant and therefore we never even start the agent recording.

### Checklist

- [ ] Updated documentation


dremin (approved):
LGTM!

PR #456, opened 2023-12-15
Minor fixes to contact history

dremin:
### Summary

- Fix blank caller ID if the feature is disabled or the user has not yet selected one
- Fix misleading UI when worker is offline and tries to place a call

### Checklist


trogers-twilio (approved):
These changes look good. Good idea checking if the worker activity is Offline to disable the click to dial feature.

PR #461, opened 2024-01-03
Fix error with no repo vars

dremin:
### Summary

Simple script change to skip generating a harmless but misleading error when no repo vars (which are not required by the template) are present.

### Checklist


trogers-twilio (approved):
Simple, but effective. Thank you for cleaning up this error.

PR #462, opened 2024-01-03
Replace tzdata with native browser method

dremin:
### Summary

The tzdata package was used for listing time zones in the schedule manager. Browsers now have a native method to return this list, so we do not need to use this dependency any longer. Saves about 100 KB.

### Checklist


trogers-twilio (approved):
Nice optimization to reduce dependencies and build size.

PR #482, opened 2024-01-29
Environment setup improvements

dremin:
### Summary

Minor improvements:
- The placeholder string is now declared as a constant so it's easier to change
- Improved behavior when a parent item cannot be fetched (this will help with the unclear errors seen in #480)
- Removed TR workspace from Terraform definition as it was only used for the workspace SID, and instead use the existing environment setup to place the SID within tfvars. This further reduces the risk of Terraform deploys, and centralizes the workspace name configuration (previously if the workspace name had been changed, you'd need to edit both the `scripts/config/mappings.json` file as well as the Terraform config)
  - This required a bump of the state version to prevent Terraform attempting to delete the workspace
- Clarify TR workspace name assumption in the docs
- Refactor the CLI fetching portion of the setup scripts to remove duplication, which makes it easier to support new resource types

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Thank you for these enhancements!

PR #485, opened 2024-01-31
Add some error handling to agent-automation

dremin:
### Summary

Add some error handling to the SelectTask and AcceptTask invocations. This is necessary because Flex throws an error trying to accept a task that has already been accepted (which happens with multiple instances open)

### Checklist


trogers-twilio (approved):
Good catch. I'm hoping we'll eventually make Single Session Guard default for all accounts so these multiple Flex session bugs will be avoided entirely. But glad to see the error handling regardless.

PR #490, opened 2024-02-07
Add eslint rule for constructors

dremin:
### Summary

If a class is exported as a singleton, and it has a constructor, and the class is imported within a flex-hook, then the constructor executes even if the feature is disabled. This eslint rule warns developers about this to prevent issues.

### Checklist


Romanti-Ezer (approved):
LGTM! 🎉 

If I remove the `// eslint-disable-next-line no-restricted-syntax` for a class with a constructor, I start seeing the error:

Screenshot 2024-03-14 at 18 16 25

Screenshot 2024-03-14 at 18 16 40

PR #495, opened 2024-02-12
Remove queue stats serverless

dremin:
### Summary

This functionality has been relocated to https://github.com/twilio-professional-services/queue-stats-and-metadata

### Checklist


trogers-twilio:
@dremin Just for completeness in documenting the reason for this change, why was this feature moved to a standalone repository versus remaining a part of the Flex Project Template?

dremin:
> @dremin Just for completeness in documenting the reason for this change, why was this feature moved to a standalone repository versus remaining a part of the Flex Project Template?

Sure! This was moved because it is merely a demonstration--it is not a complete solution, is not well documented, and not applied anywhere within the template. In addition, it is questionable whether Sync is a scalable-enough storage solution for the proposed use case.

trogers-twilio:
> > @dremin Just for completeness in documenting the reason for this change, why was this feature moved to a standalone repository versus remaining a part of the Flex Project Template?
> 
> Sure! This was moved because it is merely a demonstration--it is not a complete solution, is not well documented, and not applied anywhere within the template. In addition, it is questionable whether Sync is a scalable-enough storage solution for the proposed use case.

Got it, that makes perfect sense, thank you for the additional detail.

trogers-twilio (approved):

PR #497, opened 2024-02-14
Add branding feature

dremin:
### Summary

The branding feature aims to make it easy to apply custom branding to Flex, including a custom logo and custom colors. This is mostly just providing a nice abstraction over the Flex theme interface so that no theme code needs to be written for the majority of use cases.

- Ability to set logo by URL, including with dynamic reference of the serverless URL and/or worker attributes
  -  As part of this, refactored replaceStringAttributes as a common helper util
- Ability to easily set header and side nav colors
- Ability to set arbitrary componentThemeOverrides via config

I did not include customization of Paste tokens within this feature. I found that Flex will allow invalid values here to cause crashing in MUI code, so that does not seem safe to implement.

### Checklist

- [x] Updated documentation


dahamilton-twilio (approved):
lgtm

PR #498, opened 2024-02-15
Remove legacy icons

dremin:
### Summary

- Remove legacy search icon from the custom transfer directory because we don't need to match the OOB directory any more (now that we replace both tabs). Plus, it didn't support dark mode and used an unstable CSS class.
- Remove sample icons that have never been used (I actually thought we removed these a long time ago!)

### Checklist

- [x] Updated documentation


dahamilton-twilio (approved):
lgtm

PR #501, opened 2024-02-21
Perform flex-config env var replacement only during deploy

dremin:
### Summary

In the current state, only the environment-specific ui_attributes files in the flex-config package are parsed for environment variable replacement (such as serverless domains or object SIDs). This means that when customers also have manually-populated configs within those env-specific files, the variable replacement placeholders needs to be duplicated across each environment-specific file. Customer feedback indicated this is confusing, so the goal of this PR is to move those variable replacement placeholders to the common config file.

The setup script currently works by copying an 'example' file, performing replacement, and saving it as an env-specific file. This won't work for the common config file, since it isn't generated based on an example file. We also don't want to accidentally commit the 'filled' values, which would be easy to do with the common config.

The following changes have been made to how this works:
- Added a new env var replacement method "by string" rather than "by path", which takes the file contents as input rather than a file path, and returns the populated string as output rather than creating a file.
- Remove processing the ui_attributes flex-config files from the setup script
- Adjust the appConfig.js generation to use the new "by string" variable replacement method, since the flex-config files its config is sourced from will no longer be already filled
- Changed the flex-config deploy script to perform variable replacement at deploy-time, rather than install-time, and use the "by string" method so that no files are changed

### Checklist

- [x] Updated documentation


Romanti-Ezer (approved):
LGTM! Great job @dremin! 🚀 

I tested deploying from my personal repo to my instance, at it's working fine. I also checked the deployed Flex config ✅ 

Screenshot 2024-02-23 at 18 06 35

PR #504, opened 2024-02-22
Validate API key

dremin:
### Summary

Before we deploy anything, we need to make sure that we are on a Flex account, and that the provided API key and secret are for the same account as the provided account SID. This can help prevent accidental deployments to the wrong account when setting up an environment.

### Checklist

- [x] Updated documentation

### DIFF
--- a/.github/workflows/flex_deploy.yaml
+++ b/.github/workflows/flex_deploy.yaml
@@ -71,7 +71,7 @@ jobs:
         run: |
           npm install --ignore-scripts
       - name: validate api key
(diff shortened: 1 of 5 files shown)


dremin:
> @dremin quick question - so if the _validate-flex-account_ job fails (process exists with non-zero), the jobs that depends/needs this step (e.g. _perform-initial-serverless-release_) will not run, correct?

Yup! Here is an example run that shows what happens: https://github.com/twilio-professional-services/flex-project-template/actions/runs/7997394623/job/21841779364

PR #505, opened 2024-02-27
Extended wrap-up and other agent-automation improvements

dremin:
### Summary

Adds the following improvements to the agent-automation feature:
- Extended wrap-up
  - Two configuration options:
    - `allow_extended_wrapup`: Controls whether or not the agent has the ability to request extended wrap-up. When this is enabled, there is a toggle button added to `TaskCanvasHeader` for the agent.
    - `extended_wrapup_time`: This controls the duration allowed for extended wrap-up. This is the time that gets added to the auto-wrap timer when the agent requests extended wrap-up. If this is set to `0`, there is infinite time allowed for extended wrap-up.
  - When extended wrap-up is toggled by the agent, a new `ExtendWrapUp` action is invoked, which allows for customization around the behavior. For example, this could be used to set a task attribute for reporting purposes, or by other features such as `activity-reservation-handler` (see below).
- Auto wrap-up countdown timer: Replaces the `header` component in `TaskCanvasHeader` to show a countdown of remaining wrap-up time before auto-wrap. If the agent has requested extended wrap-up and it is configured for infinite time, then it reverts to the standard behavior of counting time upwards.
- Allow using worker attributes as part of config qualification in addition to task attributes

The `activity-reservation-handler` feature has been enhanced to support the extended wrap-up functionality by listening for the `afterExtendWrapUp` action. It will set the agent to the new configurable `extendedWrapup` activity automatically when this action is invoked.

### Checklist

- [x] Updated documentation


dahamilton-twilio (approved):
lgtm

PR #507, opened 2024-03-04
Deprecate device-manager feature

dremin:
### Summary

Flex UI 2.6 introduced a native device manager feature based on the template's device-manager feature with full feature parity (localization issue was addressed in 2.6.1).

- Applied device-manager feature translations to the native version in the localization feature (though we are missing translations for some of the additional strings in the native feature)
- Disabled device-manager by default
- Added deprecation notice to the docs
- Updated plugin deps to 2.6.1

### Checklist

- [x] Updated documentation


dahamilton-twilio (approved):
lgtm

PR #509, opened 2024-03-12
Contacts feature

dremin:
### Summary

This PR converts the `contact-history` feature into the `contacts` feature. This feature adds a contacts directory to Flex. The contacts directory consists of recent contacts, personal contacts, and shared contacts.

- Recent contacts are populated based on completed voice and chat tasks, and allow for an agent to easily redial that contact. Recents are persisted for 14 days by default, and agents can choose to clear their recents. The recents functionality was moved from local storage to Sync, so it persists across browsers correctly now.
- Personal contacts are contacts which are created by the worker, and are specific to that worker. The worker can add, modify, or delete any of their personal contacts.
- Shared contacts are contacts which are visible to all workers. By default, only workers with the admin or supervisor role can add, modify, or delete shared contacts. However, these can be optionally made editable for agents as well.

Contacts can be viewed, managed, and dialed using the contacts view added to Flex, and they are stored in Sync. In addition, a "Call Contact" section is added to the outbound dialer panel, allowing easy dialing of contacts from any view. Also, if the `custom-transfer-directory` feature is enabled, contacts are available in the transfer panel for cold or warm transfer.

The feature is now enabled by default.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Looks great! Thank you for the adding the Search enhancement and working to improve performance of large contact lists. Will be a valuable feature for our customers to take advantage of.

trogers-twilio (approved):
The latest changes around pagination and search enhancements look good. Thank you for the additional effort on these improvements!

PR #512, opened 2024-03-15
FEATURE: Post call survey

chaosloth:
### Summary

Adds post call survey UI designer, rules editor and dynamic IVR

### Checklist

- [X] Tested changes end to end
- [X] Updated documentation
- [X] Requested one or more reviewers


chaosloth:
Reviewed

chaosloth:
@dremin @JasonLearmouth @jared-hunter Hey fellas, would you mind making a review of the above

dremin:
> 202 ESLint error(s) and 6 ESLint warning(s) found in pull request changed files.

@chaosloth Could you please resolve the eslint issues? You can see them locally using `npm run lint`, or have it auto-fix as much as possible using `npm run lint:fix`.

dremin:
@chaosloth I've been playing with this and it's very nice! Good work! :tada:

However, I'm concerned about the scalability of using Sync for end-customer-facing solutions like this. Within the context of Flex UI itself Sync is usually fine, but once you add that to something that potentially runs multiple times for every customer call, the limits creep up quickly. I would be more comfortable with this if the configuration was stored in serverless, as in schedule-manager or inbound-sip-router.

Another thing I noticed is that this feature is enabled by default, but it requires manual configuration to be usable. The Terraform package needs to be updated to deploy the necessary TR workflow, and the `mappings.json` configuration should be updated to auto-populate the SID within `.env.example` based on workflow name.

I also think a "how it works" section for the docs would be helpful so that new users can understand it better. It should explain how the call gets routed to the dynamic IVR, and how the attributes are populated & can be used for Flex Insights consumption.

dremin:
Closing this PR as it has been stalled for some time. Feel free to open a new PR once the changes requested are ready!

PR #525, opened 2024-03-26
[SE-15] Add audit logging utility, add admin-ui audit log

dremin:
### Summary

Adds a simple audit event logging utility that features can use to provide an audit log. Audit logs are saved as Sync maps named per-feature. The items saved to the log have a TTL configured per the `audit_log_ttl` common config setting, which can be set via admin-ui / flex-config.

Updated the admin-ui feature to utilize the new audit logging util to save events for worker config and global config changes.

### Checklist

- [x] Updated documentation

### DIFF
--- a/docs/docs/building/template-utilities/audit-logging.md
+++ b/docs/docs/building/template-utilities/audit-logging.md
@@ -4,7 +4,7 @@ title: Audit logging
 
 Location: `plugin-flex-ts-template-v2/src/utils/helpers/AuditHelper.ts`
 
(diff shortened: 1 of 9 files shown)


trogers-twilio:
@dremin Would it be helpful to include a note in the audit logs on how to view the Sync Map data in the Twilio Console, as well as a link to the Sync Map API docs? Some customers new to Sync may not be aware of how to view / fetch the data once it's logged.

trogers-twilio:
@dremin What are your thoughts on using a [Sync List](https://www.twilio.com/docs/sync/api/list-resource) instead of Sync Map, in order to retain order of the events and perhaps make it more predictable to retrieve the most recent audit items from the list?

dremin:
> @dremin Would it be helpful to include a note in the audit logs on how to view the Sync Map data in the Twilio Console, as well as a link to the Sync Map API docs? Some customers new to Sync may not be aware of how to view / fetch the data once it's logged.

Good idea, done!

> @dremin What are your thoughts on using a [Sync List](https://www.twilio.com/docs/sync/api/list-resource) instead of Sync Map, in order to retain order of the events and perhaps make it more predictable to retrieve the most recent audit items from the list?

I can't think of a good reason not to do this. Switched to lists!

trogers-twilio (approved):
Love this idea of an audit log which can be used across the entire template as needed. Thank you for making the suggested adjustments. Looks good, fantastic work!

PR #530, opened 2024-04-03
🎨 Introducing Conversations Helper | Reduce duplicated code

Romanti-Ezer:
### Summary

Upstreaming an improvement we did while working on a customer project. We needed the `countOfOutstandingInvitesForConversation` function in a new custom feature and it wouldn't be a good practice to just duplicate that code - I also saw this was being used in Park Interactions - , so we introduced the `ConversationsHelper` to centralize this and any other future utils related to Conversations.

We also copied the types to the `ConversationState` file so future features/utils can reference types from there. (We didn't completely move from `conversations-transfer` because it was too much trouble and there were conflicts with existing types in ConversationState)

I tested `conversations-transfer` locally to make sure things are still working as expected. Park interaction we cannot test locally so I couldn't check, but it should be fine.

https://github.com/twilio-professional-services/flex-project-template/assets/31014910/25d0fd28-5de5-4f8a-8859-ad82ccfd3d3b

### Checklist


dremin (approved):
🎉

PR #531, opened 2024-04-03
Change library and refactor send-audio-rec-file feature

### LINKED ISSUE #524: Audios have poor quality when captured for sending on WhatsApp
We are using the send-audio-rec-file plugin, from [flex-project-template](https://www.twilio.com/en-us/blog/introduzindo-the-flex-project-template-developed- by - twilio-professional-services) for sending WhatsApp audio messages. At certain times, some of the operators' audio is randomly of low quality (slow, robotic voice or choppy sound).

The problem can be replicated as follows: while the operator records the audio to send, the customer must send messages.

dremin:
### Summary

- Fixes #524 by switching to an audio recording library that uses a worker instead of the main thread
- Misc refactoring around the audio recorder panel component

### Checklist

### DIFF
--- a/plugin-flex-ts-template-v2/package-lock.json
+++ b/plugin-flex-ts-template-v2/package-lock.json
@@ -16,12 +16,12 @@
         "emoji-mart": "^5.5.2",
         "lodash": "^4.17.20",
         "luxon": "^3.1.1",
-        "mic-recorder-to-mp3": "^2.2.2",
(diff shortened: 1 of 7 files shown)


trogers-twilio (approved):
Looks great! Nice work finding a package to handle the audio encoding in a separate worker thread to avoid resource competition.

PR #537, opened 2024-04-12
Refactor ConferenceService as common util, support custom-hold-music with conference feature

dremin:
### Summary

- ConferenceService actually consisted of all common functions, and other features had to either duplicate it or directly reference it. These services have been combined into a common ProgrammableVoiceService just like the other common serverless classes.
- Enhanced the new ProgrammableVoiceService to accept a custom holdUrl to be set by features
- Updated the custom-hold-music feature to set the ProgrammableVoiceService holdUrl so now when that feature is enabled, other features which use the hold-participant serverless functions also get the custom hold experience

### Checklist


plugin-flex-ts-template-v2/src/feature-library/custom-hold-music/flex-hooks/events/pluginsInitialized.ts
```diff
@@ -0,0 +1,10 @@
+import { FlexEvent } from '../../../../types/feature-loader';
+import { getHoldMusicUrl } from '../../config';
+import ProgrammableVoiceService from '../../../../utils/serverless/ProgrammableVoice/ProgrammableVoiceService';
+
+export const eventName = FlexEvent.pluginsInitialized;
+export const eventHook = function setCommonPVServiceHoldUrl() {
+  // Some features may abort the HoldParticipant action and instead call the serverless service to hold instead
+  // This makes the serverless service also use the custom hold URL
+  ProgrammableVoiceService.holdUrl = getHoldMusicUrl();
```
trogers-twilio, line 9: Just noting this as a possible improvement in a future update, not something to be addressed immediately. Would be preferable to update class properties through a setter instead of modifying the property directly. This could enable things like logging a warning if it's being set in multiple places and the developer isn't aware they're overwriting it.

trogers-twilio (approved):

PR #542, opened 2024-04-17
Make special characters in environment name work

dremin:
### Summary

Now any character that GitHub supports in the environment name can be used, except for slashes. Also enhanced the validation script to check for bad characters (slashes) before attempting deployment.

Didn't modify the documentation as I don't think we should necessarily promote the usage of special characters.

### Checklist

### DIFF
--- a/.github/workflows/flex_deploy.yaml
+++ b/.github/workflows/flex_deploy.yaml
@@ -55,6 +55,8 @@ jobs:
   validate-environment:
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
+    env:
+      ENVIRONMENT: ${{ inputs.environment }}
     steps:
       - uses: actions/checkout@v4
       - name: setup node.js
--- a/.github/workflows/flex_deploy.yaml
+++ b/.github/workflows/flex_deploy.yaml
@@ -52,7 +52,7 @@ concurrency:
   group: flex-deploy-${{ inputs.environment || github.event.inputs.environment }}
 
 jobs:
-  validate-flex-account:
+  validate-environment:
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
     steps:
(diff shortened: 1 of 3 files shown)


dremin:
> @dremin The code changes look good. Is there any concern about OS specific differences for handling of files with special characters? Was this tested on Mac OS only, or also on Windows?

This was tested on Linux and macOS only, which only restrict the `/` character. Windows has a much longer list of restricted characters: `" * / : < > ? \ |`. Do you think that we should restrict these for everyone? Restricting them only for Windows users wouldn't really work out since CI, which performs the validation, runs in Linux.

trogers-twilio:
> This was tested on Linux and macOS only, which only restrict the `/` character. Windows has a much longer list of restricted characters: `" * / : < > ? \ |`. Do you think that we should restrict these for everyone? Restricting them only for Windows users wouldn't really work out since CI, which performs the validation, runs in Linux.

I'm thinking we should be OK. A developer using Windows will be restricted by the OS anyway from creating an environment file with any of these restricted characters, so our script shouldn't need to account for it. Does that sound correct to you as well?

dremin:
> @dremin The code changes look good. Is there any concern about OS specific differences for handling of files with special characters? Was this tested on Mac OS only, or also on Windows?

This was tested on Linux and macOS only, which only restrict the `/` character. Windows has a much longer list of restricted characters: `" * / : < > ? \ |`. Do you think that we should restrict these for everyone? Restricting them only for Windows users wouldn't really work out since CI, which performs the validation, runs in Linux.

> > This was tested on Linux and macOS only, which only restrict the `/` character. Windows has a much longer list of restricted characters: `" * / : < > ? \ |`. Do you think that we should restrict these for everyone? Restricting them only for Windows users wouldn't really work out since CI, which performs the validation, runs in Linux.
> 
> I'm thinking we should be OK. A developer using Windows will be restricted by the OS anyway from creating an environment file with any of these restricted characters, so our script shouldn't need to account for it. Does that sound correct to you as well?

That's correct!

trogers-twilio (approved):

PR #548, opened 2024-04-22
Improve auto-wrap reliability, extend wrap up button fixes

dremin:
### Summary

- Fixed text wrapping inside button with a small tasks canvas
- Fixed button appearing when auto wrap is disabled but extended wrap is enabled
- Attempting to improve auto-wrap reliability by moving the logic from a component to events. The component-based approach originally seemed like the more robust solution, but it is proving to be less reliable due to the lack of guarantee that the component will actually be mounted.

### Checklist


trogers-twilio (approved):
@dremin Looks good. Thank you for making it more robust by moving the timer logic outside of the React component. Whenever possible, I prefer timers and listeners to live outside React components for the reason you highlighted, avoiding the need to be concerned with the mounted state of the component.

PR #549, opened 2024-04-24
Fix error viewing chat in Flex Insights

dremin:
### Summary

For some reason MessageInput and MessageInputActions are rendered in some chats in Flex Insights, so the canned responses menu would also attempt to render.

### Checklist


Romanti-Ezer (approved):
Tested and approved this change in Dasa!
After this fix, chats that are still open were being rendered correctly, without breaking the Insights UI

PR #550, opened 2024-04-24
[SE-22] Run and use Flex Insights proxy

dremin:
### Summary

This change fixes a few problems when running locally:
- Fixes the error banner regarding Insights being unable to load
- Allows accessing Insights from the local plugin so that testing can be performed. For example, when customizing messaging components, it can be useful to test them when opening a chat from Insights, as there is no task object.

### Checklist

- [x] Updated documentation


Romanti-Ezer (approved):
LGTM!
Great addition Sam 🎉 

image

PR #553, opened 2024-05-06
Add ability to exclude tasks from dual-channel recording

dremin:
### Summary

Many customers have compliance requirements that prevent them from recording certain calls. This PR adds configuration options to the dual-channel recording feature which will exclude a call from being recorded using the following criteria:

- Excluded attributes: Array of key/value pairs of task attributes that if present will prevent recording. For example, this could be the "direction" attribute to exclude calls that are in a specific direction, or an attribute can be set from the IVR to prevent the recording.
- Excluded queues: String array of queue names or queue SIDs that will not be recorded.

If any of the exclusions match the task, then the call will not be recorded by the dual-channel recording feature.

### Checklist

- [x] Updated documentation

### DIFF
--- a/docs/docs/feature-library/dual-channel-recording.md
+++ b/docs/docs/feature-library/dual-channel-recording.md
@@ -33,11 +33,16 @@ There are various ways to enable call recordings with Twilio Flex. Let's outline
 
 ## setup and dependencies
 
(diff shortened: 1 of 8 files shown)


dahamilton-twilio (approved):
lgtm

PR #559, opened 2024-05-29
Disable focus workaround in Flex UI >= 2.7.1

dremin:
### Summary

The message focus bug was fixed in Flex UI 2.7.1 by FLXAGNTUI-93. Disable the workaround for 2.7.1 or later.

### Checklist


dahamilton-twilio (approved):
In theory the version check could fail with older versions and > 9 patches but I don't see that being an issue in practice. Semver lib could be considered as a pattern if we need to do version checks in the future?

But LGTM

dremin:
> In theory the version check could fail with older versions and > 9 patches but I don't see that being an issue in practice. Semver lib could be considered as a pattern if we need to do version checks in the future?
> 
> But LGTM

Agreed, we should probably have a common template util for version checks that uses semver. I can get this added as a separate PR.

PR #560, opened 2024-05-29
Enable usage with the stage environment

dremin:
### Summary

Enables deploying the template to an account in stage, and fixes up areas that needed a region passed to them.

To use a stage account, create a GitHub environment variable named `TWILIO_REGION` and set it to `stage`.

Notes:
- The serverless-schedule-manager's fetch-config function is unreachable by CI due to it not being on the engineering VPN, so the serverless-schedule-manager deploy will only succeed the first time when deploying to stage.
  - The same affects the tfstate fetch function. The first terraform deploy will succeed; subsequent will fail unless the delete workflow is run.
- I was unable to test voice functionality, such as multi-call, but upon code inspection it looks like the options that need to be set will already be passed.
- To run locally against a stage account, two requirements must be met:
  - The Twilio CLI profile is created with `--region=stage`
  - The appConfig.js needs the stage URLs set. An example is available in the flex-ui-sample package.
- Within stage serverless, by default `TWILIO_REGION=stage-us1` instead of `stage`. As a result this means you need to edit `serverless-functions/.env.example` to include `TWILIO_REGION=<YOUR_TWILIO_REGION>` before deploy. The same change also needs to be done within the schedule manager serverless if you're using it.

### Checklist


dremin:
Fetching the tfstate serverless service isn't working in stage; will fix that up.

dremin:
I realized fixing the tfstate domain would be pointless since CI won't be able to fetch from the domain without VPN anyway. I ended up fixing the delete state script, which works around the issue.

Romanti-Ezer (approved):
LGTM in terms of code changes!
I see you're already successfully deploying using `TWILIO_REGION`, but if you'd like someone else double checking, I'd ask for a quick call to help me test (probably in your stage account?)

PR #561, opened 2024-05-30
Simplify and remove deprecated methods when rejecting an internal call

dremin:
### Summary

For some reason, the internal-call feature was designed so that when the receiving agent rejects an internal call, the task is accepted then immediately completed. This PR simplifies the process to cancel the task as part of the necessary cleanup serverless function.

@trogers-twilio I git-blame'd this back to the [original commit](https://github.com/twilio-professional-services/flex-dialpad-addon-plugin/blame/8d72c1e29c6f7d5bfbadcce7c1bdb6295e179224/src/customActions/internalCall/index.js#L33) 4 years ago. Do you have an idea as to why this might have been implemented this way? AFAICT it was not necessary but I could be missing something.

### Checklist


trogers-twilio (approved):
This looks good. Thank you for adding the logic to cancel the inbound task to avoid a potential race condition with the "conference-end" event handler.

PR #562, opened 2024-05-31
Improve pause-recording robustness

dremin:
### Summary

- Handle multiple simultaneous recordings (dual-channel and conference recordings at the same time)
- Handle missing recording (instead of showing a failure message, such as when recording is completely disabled, or excluded by rule)
- Improve logs

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
This looks good. Great idea to support pausing both conference recording and call recording if both happen to be enabled.

PR #572, opened 2024-06-12
Update TS linting packages

dremin:
### Summary

The eslint-config-twilio-ts package is unmaintained and was using old incompatible TS linting packages. Therefore, I have switched us to use the base eslint-config-twilio package and added back the TS rules from the other package, accounting for changes in the rules since that one was last updated.

This fixes the version warning when running the linter.

### Checklist


Romanti-Ezer (approved):
LGTM!
No errors running the updated eslint locally either ✅

PR #573, opened 2024-06-12
[SE-21] Additional validations before deployment

dremin:
### Summary

Updates the validation script that runs prior to deployment to improve output and validate the following additional items:

- All necessary environment variables are present
- The deployed Flex UI version is supported by the template (as documented, we support currently-maintained versions of Flex UI 2.x only)
- The configured Flex TaskRouter workspace is named the same as the value configured in `scripts/config/mappings.json` so that the setup script is able to successfully obtain resource SIDs

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
This looks great! Love the additional validation to catch potential issues up front and the clear logging if problems are found.

PR #576, opened 2024-06-17
[SE-17] Revise park timeout for SMS

dremin:
### Summary

WhatsApp and Messenger have a 24-hour maximum after the last message sent, and web clients also have their own timeout. However, SMS conversations work great past 24 hours, so let's remove the timeout for those only.

### Checklist

- [x] Updated documentation


Romanti-Ezer (approved):
LGTM!

PR #584, opened 2024-06-27
Fix crash with missing dispositions config props

dremin:
### Summary

If global attributes arrays are not present in the config, flex-ui will crash. This fixes that.

### Checklist


trogers-twilio (approved):
Thank you for finding and fixing this bug

PR #590, opened 2024-07-03
Check dialing permissions during number validation

### LINKED ISSUE #580: Cold Transfer Failing without error handling
Using the custom transfer directory, cold transferred calls that are to destinations that do not have voice geo-permissions enabled are failing to connect. The agent is left with the impression that the cold transfer was successful, however the end-user is disconnected.

The issue is that geo-permissions are applied correctly but the `functions/common/twilio-wrappers/programmable-voice.private.js` blindly executes the transfer without checking if an error (such as `[13227]`) has occurred and return that error to the Flex UI.

# Steps to replicate:
- Enable geo permissions for a given country
- In Flex receive or place a call
- Attempt to cold-transfer said call to this blocked country

dremin:
### Summary

Fixes #580

### Checklist


djayanna (approved):
Looks good to me. One thing I was pondering is, since the function now only returns the result from permission checks (bool and invalid reason) from lookup, which works for this use case, does it make sense to pass permissions to validate to the function? Also, this is coming from the fact that I'm not well-versed with the Lookup API and dialing permissions. :-)

PR #598, opened 2024-07-16
Create helper util for getting Flex feature flags

dremin:
### Summary

There are several places in the plugin where we need to check whether certain Flex feature flags are enabled. Typically these are beta features, so they may not be enabled in every account or customer environment. These flags can typically also be overridden within the plugin's appConfig.js when running locally.

This PR adds a helper util for getting the enabled state of a Flex feature flag. This considers both the account setting and any local overrides. Previous code did not consider localOverrides, which could put Flex in an inconsistent state.

### Checklist


Romanti-Ezer (approved):
Great job with this helper Sam! 🎉 
Tested and explored locally, LGTM! ✅ 

Screenshot 2024-07-25 at 14 42 36

PR #601, opened 2024-07-22
[SE-43] Add and use version utility, fix 2.8.x button sizes

dremin:
### Summary

Adds a semver-based version-checking utility to the template to make it easier for features to validate supported Flex UI versions.

Refactored existing version-checking code to use this new utility.

Added version checks for task canvas header buttons so that they use the correct size in 2.8.x, which was increased.

Note: The semver package was already included in our plugin bundle due to other dependencies pulling it in; this should not increase plugin size.

### Checklist


Romanti-Ezer (approved):
LGTM!
I did some debugging locally and it seems to be working as expected ✅ 

image

PR #605, opened 2024-07-24
Supervisor barge/coach optimizations

dremin:
### Summary

Made changes to reduce Redux and Sync updates:
- Moved sync doc subscription from coaching status panel component to pluginsInitialized hook so that it doesn’t resubscribe for every task
- Skip updating Redux state on doc update if supervisor array is empty before and after the update
- Unmount coaching status panel and assistance button when wrapping
- Stop setting the subscription flags to false at wrapup, as it isnt unsubscribing
- Update assistance doc only if a change was made

@aestellwag could you please help validate that I'm not doing anything here that has failed in the past?

### Checklist


dremin:
This is going to conflict with #610 so I'll probably combine them.

PR #608, opened 2024-07-30
Avoid unnecessary conference state updates

dremin:
### Summary

The ParticipantActionsButtons component was creating state for every conference participant unnecessarily. Because SetComponentState merges updates, the state would grow quite large. Changed this to only set state when needed (when the kick buttons are clicked).

Removed code that tried to clean up the state, which doesn't work due to SetComponentState's merge logic.

### Checklist


trogers-twilio (approved):
Much cleaner, thank you for catching and fixing this memory leak

PR #610, opened 2024-07-31
Fix misc supervisor barge/coach issues, improve UI, improve agent assistance (SE-47)

dremin:
### Summary

- Reduced UI confusion by using a segmented control for the supervisor
- Use built-in Flex state for enabling/disabling buttons instead of trying to manage it ourselves:
  - Fixes buttons staying enabled indefinitely when call ends while monitoring
  - Fixes buttons staying enabled when selecting another call in teams view while monitoring
- Made changes to reduce Redux and Sync updates (originally #605):
  - Moved sync doc subscription from coaching status panel component to pluginsInitialized hook so that it doesn’t resubscribe for every task
  - Skip updating Redux state on doc update if supervisor array is empty before and after the update
  - Unmount coaching status panel and assistance button when wrapping
  - Stop setting the subscription flags to false at wrapup, as it isnt unsubscribing
  - Update assistance doc only if a change was made
- Refactored reusable barge/coach logic out of the button component and into a helper
- Refactored refresh logic out of the components flex-hooks and into the pluginsInitialized event
- Stopped using stickyWorker from Flex state as it required always going to the Teams view after reload. This fixes supervisors always refreshing into the Teams view instead of wherever they were
- Fixed inconsistency where coaching status would show to agent if the supervisor first coached then barged, but not if they simply barged
- Fixed overloaded syncSubscribed state
- Fixed overloaded supervisorArray state
- Fixed private mode button unexpectedly adding to doc when not monitoring
- By switching to taskWrapup event, fixed wrapup handler not running if Flex UI refreshed during call
- Supervisor monitor panel now properly unsubscribes
- Converted state to slice
- Misc UI improvements
- Misc optimizations
- Enhanced agent assistance capabilities (SE-47)
  - Added button for supervisor to go directly to the task in the Teams view (if the worker isnt filtered out)
  - Added ability for the supervisor to clear the assistance request by clicking the icon in Teams, with confirmation modal
  - Fixed notification issues with multiple agents requesting assistance (now a notification is shown for each agent instead of just the first)
  - Agent interface now temporarily subscribes to the assistance doc when requested so that state is updated if the supervisor clears the request
  - Moved supervisor agent assistance doc subscription logic out of component into pluginsInitialized and removed unnecessary state

### Checklist

- [x] Updated documentation


dremin:
I decided to combine all of the work I've been doing on this feature into one PR because it is all intertwined. The UI changes still need updated screenshots for the docs.

aestellwag (approved):
UI changes look very clean, I like it a lot.  I attempted to break this starting, ending calls, refreshing in the middle of coaching, barge, monitor.  Also refresh while agent assistance was up, as well as privacy and notification on and off from the supervisor.  The state cleaned itself properly and restored itself to the last known state.

Looks good, no errors encountered while testing with many many browser refreshes.

PR #618, opened 2024-08-23
Update to use Flex 2.8.x locally

dremin:
### Summary

As it says on the label!

### Checklist


trogers-twilio (approved):
Thank you for keeping the Flex UI version up to date

PR #623, opened 2024-09-03
Update sip-support for Flex UI 2.7+

dremin:
### Summary

Flex UI 2.7 and 2.8 call WrapupTask instead of HangupCall when the end call button is clicked and the embedded voice client is not used. See TECHFLEX-181

### Checklist


trogers-twilio (approved):
This looks great! Thank you for adding the dynamic Flex UI version check. Once 2.7 is the oldest supported Flex UI release, we can remove this HangupCall hook entirely in additional template releases.

PR #624, opened 2024-09-03
[SE-50] Deprecate external directory config option

dremin:
### Summary

Given that the contacts feature is the scalable way to save contacts, and that the custom-transfer-directory integrates with the contacts feature, this deprecates the flex-config-based external directory entry support in custom-transfer-directory.

The functionality is still there in case a customer using it updates their copy of the template. A warning log has been added in that case.

### Checklist

- [x] Updated documentation

### DIFF
--- a/docs/docs/feature-library/custom-transfer-directory.md
+++ b/docs/docs/feature-library/custom-transfer-directory.md
@@ -17,9 +17,7 @@ This feature enables the replacement of the queue and worker transfer directorie
 
 It also enables the addition of an external directory, enabling the following behavior
 
-- present a list of external transfer numbers
-  - includes contacts from the `contacts` feature if enabled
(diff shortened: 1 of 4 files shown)


trogers-twilio:
@dremin Does the "contacts" feature support toggling cold or warm transfer per record like the Flex config external directory does? I didn't see that option in the "contacts" feature docs, but I may be missing it.

dremin:
> @dremin Does the "contacts" feature support toggling cold or warm transfer per record like the Flex config external directory does? I didn't see that option in the "contacts" feature docs, but I may be missing it.

You are correct, it does not.

trogers-twilio (approved):
Agreed with deprecating this Flex config custom directory option due to scalability concerns. We'll add a backlog item to enhance the "contacts" feature with the ability to enable / disable warm and cold transfer per contact.

PR #625, opened 2024-09-04
Better handle another worker participant joining

dremin:
### Summary

In Flex UI 2.7 and later, the current worker can have multiple conference participants representing themself when using the "Join Call" button. Therefore, any time we are checking for the current worker participant, we need to also check that they are joined.

Note that the dual-channel-recording feature does not currently handle this when recording the worker call leg and the worker uses the "Join Call" button. There is simply no recording after that point, since the recording is only initiated upon task acceptance. This limitation has been documented, in addition to a notice recommending the new native dual-channel conference recording capability.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Looks good. Thank you, Sam!

PR #628, opened 2024-09-09
Multi-call refactor

### LINKED ISSUE #133: Enhancement: Support more than two multi-calls
Do you think it would be difficult to support more than just two calls at a time? I was thinking it could be as simple as storing a list of calls and devices, destroying and adding as calls are coming in, but maybe there's more to it? Thanks!

dremin:
### Summary

Re-wrote the multi-call implementation around the new functionality in Voice SDK 2.11 and newer, included with Flex UI 2.8 and newer.

The new implementation is no longer limited to two simultaneous calls--up to 6 have been tested!

Fixes #133.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Fantastic work! Love that we're taking advantage of the newer Voice SDK features for multi-call handling and now supporting far more simultaneous calls.

PR #648, opened 2024-10-02
Fix outbound cold transfer when using dual-channel-recording

dremin:
### Summary

The custom-transfer-directory feature's cold transfer functionality needs a call SID to initiate the transfer for. Previously this used the call_sid attribute, as this should always be the customer call SID, however due to an issue in the dual-channel-recording feature, this is not currently the case when recording the worker leg and placing an outbound call.

We will fix dual-channel-recording separately, but this is a quick fix to use the known-good SID first.

### Checklist


dahamilton-twilio:
LGTM as a workaround.

Looks like the root cause is likely just an oversight that worker leg can be recorded

dahamilton-twilio (approved):

PR #653, opened 2024-10-11
Fix order of operations for VM task cancel reason

dremin:
### Summary

The call update was canceling the task before we had a chance to set the reason. Reversed the order of operations, and added a failsafe in the event that the task cancels but the call doesn't update.

### Checklist


dahamilton-twilio (approved):
LGTM - Nice fix, I could see this being difficult for customer to track down the timing issue.

PR #654, opened 2024-10-11
Add action to select CRM tab

dremin:
### Summary

Now features can programmatically select a specific tab in the CRM container.

Also made tabs that don’t target a task work better, and adjusted logic so that adding tabs dynamically is possible.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Looks good! Clever use of tabState and the custom action "SelectCRMContainerTab" listeners to handle the tab change.

PR #667, opened 2024-11-07
Pin emoji-mart@5.5.2

dremin:
### Summary

emoji-mart 5.6.0 cannot be built by the current Flex plugin toolchain, so let's pin to the last working version for now.

### Checklist


Romanti-Ezer (approved):
LGTM ✅ 
Thanks for pushing the fix

PR #670, opened 2024-11-15
Fix omni-channel-capacity-management (#669)

dremin:
### Summary

The serverless function that the omni-channel-capacity-management feature uses is only intended to be used by supervisors or admins. This broke the feature for agents. Updated this function to be usable by agents when a workerSid is not provided, which will modify the current worker's channel.

### Checklist


trogers-twilio (approved):
These changes look good.  Thank you!

PR #671, opened 2024-11-18
Fix multi-call auto-holding for other channels

dremin:
### Summary

For multi-channel agents with the multi-call feature enabled, previously accepting a non-voice task would auto-hold any active calls. This PR fixes that behavior so the call remains unheld when accepting a non-voice task.

### Checklist


trogers-twilio (approved):
This change looks good. Thank you!

PR #672, opened 2024-11-18
Fix flex-config deploy error code

dremin:
### Summary

If flex-config deploy failed, we were catching the error and not reporting any obvious failure in the deploy. This change causes the deploy to report a failure and also removes the duplicate logs.

### Checklist


trogers-twilio (approved):
Thank you for this change. Simple and effective.

PR #674, opened 2024-11-19
Fix conversation-transfer issues from Support

dremin:
### Summary

1. Fixed handling of emails. Each email recipient is a conversation participant, so we were erroneously showing the Leave button when the email recipient count pushed the total participant count above two. Fixed this by counting only one participant that has a binding, as we know those participants are external. Also moved this logic to ConversationsHelper so that park-interaction can benefit as well.
2. Fixed a timing hole after the agent chooses to transfer, where they could click the End button while the invite API request is inflight. This would break the task created by the transfer. Fixed this by holding a bit of state while this API request is executed, checked within WrapupTask, which is blocked if the transfer is pending.

Referenced Jiras:
- TECHFLEX-523
- TECHFLEX-582

### Checklist


trogers-twilio (approved):
Looks good, thank you! Great idea moving logic into ConversationsHelper to be re-used across features.

PR #688, opened 2024-12-10
Flex UI 2.10

dremin:
### Summary

Updated to Flex UI 2.10 and regenerated package-lock.json to avoid the dependency issue with props.

### Checklist


trogers-twilio (approved):
Thank you for addressing this dependency issue.

PR #691, opened 2024-12-11
Update root package dependencies

dremin:
### Summary

Regenerated the root-level package-lock file to update dependencies. This should also resolve #684, which I avoided merging as its changes would cause package-lock to get modified again when a user pulled down the repo and performed an npm install.

### Checklist


trogers-twilio (approved):
Thank you for this dependency update!

PR #693, opened 2024-12-11
Update to Flex plugins CLI 7.1.0

dremin:
### Summary

The main change here is that a lot of dependencies are now correctly marked as dev only. Doesn't seem to affect the production plugin bundle size at all.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Looks good. Thank you for keeping our Twilio CLI and CLI plugins up to date.

PR #694, opened 2024-12-12
Fix queues data tiles breaking on Flex UI 2.7

dremin:
### Summary

If enabled, the feature broke Flex UI 2.7 or older because the API is not available until Flex UI 2.8.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Code changes and doc updates look good. Thank you for catching this Flex UI version dependency.

PR #695, opened 2024-12-13
Add reference to Flex Plugin Library in readme

dremin:
### Summary

As discussed :)

### Checklist

- [x] Updated documentation


trogers-twilio:
Thank you Sam! Should we also direct the reader that they should remove any features from the Flex Project Template that overlap with the corresponding Plugin Library feature, linking to our guide on how to remove features from the Template?

dremin:
> @dremin Thank you Sam! Should we also direct the reader that they should remove any features from the Flex Project Template that overlap with the installed Plugin Library features to avoid Flex configuration conflicts, linking to our guide on how to remove features from the Template?

I'm more inclined to mention that the template versions should be used instead if they are going to deploy the template anyway, as there are still significant downsides to using the plugin library versions of these features. What do you think?

trogers-twilio (approved):
This looks great! Thank you for these documentation updates and encouraging Flex users to leverage the native Plugin Library features when possible.

PR #696, opened 2024-12-13
Support out-of-box localization

dremin:
### Summary

- Disable the localization feature if the native feature is enabled on the account
- Adjust `getUserLanguage()` helper to return the language set by the native feature when enabled
- Update docs (also updated the device-manager doc due to some outdated info I noticed, as well as some notes regarding TECHFLEX-462).

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Looks good, thank you for accounting for the language to return in the util function as well.

AndreyTwilio:
I guess this now could be merged?

PR #699, opened 2024-12-19
[SE-48] Support native messaging transfers

dremin:
### Summary

- Prevented conversation-transfer feature from being enabled when native messaging transfers are enabled
- Updated custom-transfer-directory feature to use the native interfaces for starting a messaging transfer when the native feature is enabled
- Updated conversation-transfer doc to list differences between it and the native feature

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
This looks great! Thank you for addressing this potential feature conflict and making the custom directory compatible with native digital transfers.

PR #701, opened 2025-01-07
Fix endless ringing in ring-notification

dremin:
### Summary

When receiving multiple tasks, and they are both alerting at the same time, we would overwrite the `mediaId` of the first task, so we could not stop it when accepting. We avoid that scenario by returning in the `play` function if we are already playing.

As an initial attempt at the fix, I tried stopping the initial media when the second `play` was invoked, however, I found that in the case multiple reservations come to you at the same time, the `stop` would not actually prevent it playing when it was called before the audio file loaded. This may be an issue with the Flex `AudioPlayerManager`, but I did not look into this much further as the solution I landed upon seems to work well.

### Checklist


trogers-twilio (approved):
Nice catch. For future enhancement, we might want the ringing to begin again if there's still a reservation in pending after the current ringing one has triggered a stop. But not something to deal with in this PR.

PR #703, opened 2025-01-08
Feature/enhanced queue filter array support

SeanHolden:
### Summary

_Provide a summary of the change_

### Checklist

- [x] Updated documentation
- [ ] Requested one or more reviewers


dremin (changes requested):
Thanks for adding this @SeanHolden! A couple issues I noticed:
1. The change removes the global queue filter inadvertently, it needs to be added back (ideally with the same enhancement!)
2. The type definitions need to be updated. For the worker attributes, this is in `types/task-router/Worker.ts`.

dremin:
Closing in favor of #815.

PR #706, opened 2025-01-27
Update TR SDK Task type

dremin:
### Summary

Allows adding event listeners on the TR SDK task object. Not used anywhere in the default feature set.

### Checklist


trogers-twilio (approved):
Thank for adding support for this scenario in case it arises

PR #719, opened 2025-02-28
Update agent assistance alert icon

dremin:
### Summary

The native alerting feature uses a bell icon as well, so we need to change to something else. Used the same help icon from the agent view for consistency. Also updated the style to match other buttons in the header.

### Checklist

- [x] Updated documentation


trogers-twilio (approved):
Thank you for catching this feature icon conflict and resolving it so quickly!

PR #722, opened 2025-03-20
Fix UI issues with number validation

dremin:
### Summary

- Add a try/catch in case the serverless request to validate number fails
- Fix missing phone number in error message

### Checklist


trogers-twilio (approved):
Thank you for making this feature more robust

PR #738, opened 2025-04-30
Fix create-task vulnerability

dremin:
### Summary

- Fix create-task function being accessible without auth
- Fix TR helpers returning token in response
- Also fixed unprotected unpark-interaction function

### Checklist


trogers-twilio (approved):
Looks good! Thank you for addressing this.

PR #748, opened 2025-05-30
Fix bad state when ending monitor leg by PSTN/SIP phone

dremin:
### Summary

Previously, we relied on the StopMonitoringCall action to clean up state and the Sync doc. However, this does not account for situations where the monitor call leg ends due to other reasons, such as hangup from a PSTN/SIP phone. For example, if the monitor call leg is ended from a SIP device, the Sync doc never gets updated, which leaves an incorrect coaching indicator on the agent's desktop.

Flex state maintains information about the monitor call leg, so move the logic to a hidden always-mounted component that listens to these state updates instead, and perform the clean up there.

### Checklist


plugin-flex-ts-template-v2/src/feature-library/supervisor-barge-coach/custom-components/CoachingStatusMonitor/CoachingStatusMonitor.tsx
```diff
@@ -0,0 +1,46 @@
...
+    // we need to update the Sync doc to remove the supervisor after they stop monitoring the call
+    if (!isAgentCoachingPanelEnabled() && !isSupervisorMonitorPanelEnabled()) return;
+    SyncDoc.initSyncDocSupervisors(monitoredAgentWorkerSid, '', myWorkerSID, '', '', 'remove');
+
+    // Remove the agent we monitored from local storage
+    localStorage.removeItem('agentWorkerSID');
+  };
+
+  useEffect(() => {
+    if (monitoringState === 2 && monitoringTask?.sid) {
```
trogers-twilio, line 30: @dremin Are the monitoringState values documented anywhere? I think this code is readable based on the note within the if block, just curious if there is any public documentation we can point to for these values in case they were ever changed.

trogers-twilio: Thanks for confirming these values are not currently publicly documented. The note in the code will be sufficient to indicate what it means.

trogers-twilio (approved):
Looks good. Thank you for enhancing this to handle these alternative scenarios of the monitored call ending.

PR #772, opened 2025-08-05
Salesforce integration example feature

dremin:
## Overview

This feature provides an enhanced Salesforce integration which replaces the out-of-box Salesforce integration plugin. It can be used as a starting point for customizing Salesforce integration functionality, which is not possible when using the out-of-box integration.

Functionality included within this implementation:
- Activity logging
  - Creates activity upon task completion or cancellation
  - Includes agent copilot disposition and summary
  - Relates the activity to the dialed or screen-popped record, or the agent-selected record
- Click-to-dial
- Screen pop
- UI enhancements
  - Disables the pop-out and pop-in buttons while on a call, to prevent accidental call hangups
  - Hides the CRM container when embedded
  - When screen pop returns multiple records, a dropdown is added to the interface for the agent to select the appropriate record for activity logging

### Context

Flex includes a Salesforce integration out-of-the-box, however, it is not fully customizable. If the out-of-box integration does not fully meet your needs, you may end up needing to build your own enhanced integration, re-creating the functionality included in the out-of-box integration.

### Objective

This `salesforce-integration` feature aims to be used as a starting point for your own customized Salesforce integration. The feature offers largely the same baseline functionality of the out-of-box integration, as well as some critical usability enhancements:

- Disables the pop-out and pop-in buttons while on a call, to prevent accidental call hangups
- When screen pop returns multiple records, a dropdown is added to the interface for the agent to select the appropriate record for activity logging

### Checklist

- [x] Updated documentation


dremin:
A few things to ponder:

1. There are some additional utility bar customizations we could apply that are explored [here](https://github.com/twilio-professional-services/plugin-salesforce-utility-item-customization). It may make sense to add them as optional items so we could retire that repo.
2. For multiple match screen pop results, we display a dropdown in the agent UI. However, customers may want to use LMS to offer a more direct UI within the Salesforce record itself. Since the template does not deploy managed objects in Salesforce, we cannot include a working out-of-box example of this. Maybe it would be worth adding a mention in the docs.

dremin:
Both items in the previous comment have now been addressed.

dremin:
Realized one more thing to add--a notification when attempting click-to-dial in the Offline activity.

PR #789, opened 2025-10-06
Fix infinite loading if no workers/queues present

dremin:
### Summary

Utilizes a new prop from Flex UI that allows us to differentiate between "loading" and "no workers/queues present".

### Checklist


trogers-twilio (approved):
Great idea leveraging the new Flex UI prop to address this edge case!

PR #848, opened 2026-02-27
Fix up Flex state when multi-call device call ends

dremin:
### Summary

The outbound dialpad button is disabled when there is an active call in Flex state. When a call handled by the Flex device ends, it is removed from state, but when a multi-call device call ends, we were not removing it from state. This means that if the last call to end was not on the Flex device, the user was prevented from placing an outbound call. This PR fixes that.

### Checklist


dremin:
Just going to merge this since it was confirmed to fix the issue.

